PR #69, opened 2015-01-24
Add elfeed-title-cleanup method

tzneal:
Adds a method for cleaning up the titles before display to the
user. This includes decoding numeric HTML entities along with a
defcustom alist for user customizable replacements in the form of
elfeed-title-cleanup-replacement-regexps.

### DIFF
--- a/elfeed-lib.el
+++ b/elfeed-lib.el
@@ -13,6 +13,14 @@
 (require 'time-date)
 (require 'url-parse)
 
+(defcustom elfeed-title-cleanup-replacement-regexps  '(("&amp;" . "&") ("&nbsp;" . " ")
+                                                       ("&lt;" . "<")  ("&gt;" . ">")
+                                                       ("&quote;" . "\""))
+  "Alist of regex and replacement values to be applied by
+`elfeed-title-cleanup' before trimming."
+  :group 'elfeed
+  :type 'alist)
+
 (defun elfeed-expose (function &rest args)
   "Return an interactive version of FUNCTION, 'exposing' it to the user."
   (lambda () (interactive) (apply function args)))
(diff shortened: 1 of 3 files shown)


skeeto:
Character entities are decoded by the XML parser, so no further decoding is necessary. If you're still seeing an entity in the display, it means that feed is double encoding entities. Neither RSS nor Atom allow HTML content inside titles, so, unlike the content of an entry, titles shouldn't go through an additional round of decoding/interpretation. There's something to be said about making *some* effort to support broken RSS feeds -- this is already done to an extent -- but doing a second entity decoding would break properly crafted feeds that intend to display an encoded entity literally (e.g. an article about a particular HTML entity).

White space cleanup, as currently done, is necessary because XML is white space preserving by default. The RSS spec has nothing to say on the matter, but the Atom spec [explicitly allows collapsing whitespace for display purposes](https://tools.ietf.org/html/rfc4287#section-3.1.1.1).

> If the value is "text", the content of the Text construct MUST NOT
> contain child elements.  Such text is intended to be presented to
> humans in a readable fashion.  Thus, Atom Processors MAY collapse
> white space (including line breaks) and display the text using
> typographic techniques such as justification and proportional fonts.

skeeto:
I forgot to ask in my initial response: can you show me what in particular you're trying to fix? I'd like to see the feed that exhibits the problem. Right now I'm just guessing about feeds double-encoding entities, but perhaps there's a real issue somewhere.

Thanks for taking the time to write a patch. I just want to make sure we fix the right thing.

tolchz:
Sure, try this feed: http://isocpp.org/blog/rss/category/news

Looks like the title is a CDATA section which is why it's not being decoded in the first pass.  

<title><![CDATA[Interlude: C++’s Strides in 2014&#8212;K-ballo]]></title>

No idea if this is valid, feedvalidator.org says it is with warnings about interoperability.

skeeto:
Yup, that's the double-encoding thing. That's why Feed Validator was
warning about interoperability. There's no way for a client to know
decode the title, other than as plain text. The XML decoder takes care
of the CDATA on the first pass, which is correct, but those HTML
entities on the second pass are ambiguous.

PR #70, opened 2015-02-05
Provide extension points to handle errors

remyhonig:
Call elfeed-handle-http-error or elfeed-handle-parse-error when a feed cannot be updated. This enables me to handle errors in https://github.com/remyhonig/elfeed-org. I would like to use "defadvice" there to mark the feed in my org tree as unretrievable so I can identify bad feeds and maybe ignore them when they fail consistently.

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -102,6 +102,18 @@ when they are first discovered."
   :group 'elfeed
   :type 'integer)
 
+(defvar elfeed-http-error-hooks nil
+  "Hooks to run when an http connection error occurs. 
+It is called with 2 arguments. 
+The first argument is the url of the failing feed. 
+The second argument is the http status code.")
+
+(defvar elfeed-parse-error-hooks nil
+  "Hooks to run when an error occurs during the parsing of a feed.
+It is called with 2 arguments. 
+The first argument is the url of the failing feed. 
+The second argument is the error message .")
+
 (defvar elfeed-connections nil
   "List of callbacks awaiting responses.")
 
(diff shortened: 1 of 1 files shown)


skeeto:
This is a great idea and I'll definitely merge this in, or at least something close to it.

My one concern is that I generally consider advice as being a last resort since it's so clunky. I don't want to design an interface to require it. I was thinking about changing those functions to `elfeed-handle-http-error-function` and `elfeed-handle-parse-error-function` variables (or something along those lines) instead, to be funcalled on error. But you probably wouldn't want elfeed-org to completely take over error handling. You just want to know when it happens.

So it really calls for a hook, perhaps prepopulated with the default printed message handler. The problem is that there's not really a convention for prepopulating a hook, Removing prepopulated entries from a hook is complicated business. It depends on whether or not Elfeed has been loaded, and whether extensions that would use it, like elfeed-org, have loaded. That means something nasty like `eval-after-load` would be needed. It's too bad there's no JavaScript-like bubbling convention that allows a hook function stop the hook invocation propagating.

Thoughts? Are there any existing projects out there that have dealt with this asynchronous error handling issue?

remyhonig:
Hi Chris, thanks for considering this. How about keeping the `message` calls in the `elfeed-update-feed` function like they are in `master` but add right after those calls a call to a new hook. This would avoid the complications of pre-populating that hook, but would enable anyone to be notified of errors. You're right, `elfeed-org` does not need to handle the errors at all, it just needs to be notified.

In javascript this type of handling is done with promises (https://www.promisejs.org/) and for elisp there is a package that can help with this as well. It is https://github.com/kiwanami/emacs-deferred . For a usage exampel look at this section https://github.com/kiwanami/emacs-deferred#deferred-combination--try-catch-finally . Using this package would require quite some refactoring though and you'd still have to create a hook where other packages can register their handling functions.

I'm willing to create a new PR with a hook if you like.

skeeto:
Ok, I like this idea. Could you modify your PR to do this? Keep those
two new functions defined like you have it. I like having them for
last-resort advice. Put the run-hooks call in those functions. I'll
merge it with that change.

Sorry, I traveled about 5,000 miles last week and was too busy to
respond earlier.

remyhonig:
No worries, I'll make the adjustment to the PR in the coming week

remyhonig:
@skeeto Fixed, let me know if you'd like me to correct anything.

skeeto:
Rebased onto master as bc860f6 and 4df2c83. Thanks!

Nice commit logs, too.

PR #76, opened 2015-03-10
elfeed-show.el: fix yank to clipboard in OSX

heikkil:
While  sexp ''(kill-new link)" is present in function elfeed-search-yank, it is missing from elfeed-show-yank. Pressing y in entry window has not been adding the article URL to the clipboard in OS X. Adding this sexp fixes it.


skeeto:
Nice catch, thanks!

PR #77, opened 2015-03-11
search-{tag,untag}-all, search-update: docstrings

jinnovation:
This commit adds docstrings to three of the more fundamental functions
of the elfeed interface. The docstrings themselves are fairly
self-explanatory, but including helps to round out the package's
presence in Emacs' built-in documentation system.


skeeto:
Thanks! Originally these were just anonymous functions assigned directly
in the keymap. Then I used defalias to give those anonymous functions
names so that they'd show up nicely in C-h m. Finally with these proper
docstrings we're almost back to normal functions.

PR #79, opened 2015-03-14
DRY out select tests

jinnovation:
Remove the code reuse from myriad tests, e.g. `elfeed-feed-type` and `elfeed-format-column`.


skeeto:
Thanks for taking the time to do this. However, I don't think DRY generally applies to unit testing. Unit tests are redundant by nature, and there's not really any reason to shy away from that. While using fancy anonymous functions reduces some of the redundancy, it makes most of the tests harder to understand.

I did accept fc09894 since that's a useful fix.

jinnovation:
That's fair; was good practice at Emacs Lisp in any case. Glad to see that at least something came of it!

Issue #83, opened 2015-04-03
[FEATURE] `elfeed-search-browse-url` ability to open externally


jinnovation:
Many times, `browse-url` is set to use text-based browser, such as `eww`, within Emacs, but there are many feeds that do not display well within Emacs. In such cases, user would then be required to call another function from within said browser to display the page in external browser. This is an extra step that can be easily circumvented by having an `elfeed-search-browse-url-external` function, possibly bound to `B`, that opens in external browser directly.

Another solution that I can think of is to redefine elfeed-search-browse-url to accept a prefix arg, in the presence of which case it'd open the URL externally.

jinnovation:
Another solution that I can think of is to redefine `elfeed-search-browse-url` to accept a prefix arg, in which case it'd open the URL externally.

skeeto:
Addressed by #84.

PR #84, opened 2015-04-03
elfeed-search-browse-url: allow generic browser open

jinnovation:
In response to Issue #83 (feature request), this commit updates
`elfeed-search-browse-url` to use the prefix argument as
`use-generic-p`, which then toggles whether or not `browse-url` or
`browse-url-generic` is used to open the URL in question.

This modification is so that, in the case where user is aware that a
given URL does not render well in the "primary" browser function defined
by `browse-url-browser-function', he/she can opt to use a second browser
-- possibly an external one, such as Chromium -- to open it.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -384,14 +384,19 @@ expression, matching against entry link, title, and feed title."
                collect it into selected
                finally (return (if ignore-region (car selected) selected))))))
 
-(defun elfeed-search-browse-url ()
-  "Visit the current entry in your browser using `browse-url'."
-  (interactive)
+(defun elfeed-search-browse-url (use-generic-p)
+  "Visit the current entry in your browser using `browse-url'.
+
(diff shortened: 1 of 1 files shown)


jinnovation:
I additionally propose in #83 to bind this new functionality to <kbd>B</kbd>, but I'll leave that for a later PR if this one gets accepted.

skeeto:
This is kind of an odd, subversive way to use browse-url, which makes me
hesitate on this patch. But I can see the usefulness of it, and it's not
intrusive, so I've merged it as 91a1498, with one tweak: the new
argument is optional so that existing calls to this function don't
break.

skeeto:
Oh yeah, and I wanted to add that the prefix argument is fine, but I don't really want to bind this directly to a key. You can easily do that yourself in your own config, though.

jinnovation:
Optional argument makes perfect sense; my fault for not catching that. Much thanks for the accept.

PR #90, opened 2015-05-29
adding dowloading all enclosures of the current entry function and binding

seblemaguer:
I propose to add a download function and the binding to easily access to enclosures (I used this method to get my podcasts audio files)

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -216,6 +216,20 @@ directory and saves all attachments in the chosen directory."
   :group 'elfeed)
 
 
+(defun elfeed-download-enclosure (url path)
+  "Download asynchronously the enclosure from URL to PATH"
+  (if (fboundp 'async-start)
+      ;; If async is available, don't hang emacs !
+      (async-start
+       (lambda ()
+         (url-copy-file url path t))
+
+       (lambda (result)
+         (message (format "%s downloaded" url))
+         )
+       )
+
+    (url-copy-file url path t)))
 
 (defun elfeed-split-ranges-to-numbers (str n)
   "Convert STR containing enclosure numbers into a list of numbers.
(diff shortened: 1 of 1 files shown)


skeeto:
I like the idea -- I'll use it myself when it's ready -- but this patch still needs more work. As is, a feed host could perform a remote code execution attack against Elfeed's Windows users with an enclosure URL along the lines of:

    http://example.com/~\.emacs

Or, with `~/Downloads` being a likely `output-directory` candidate:

    http://example.com/..\.emacs

Instead of `format`, use `expand-file-name`. Instead of `replace-regexp-in-string`, use `url-generic-parse-url`, `url-path-and-query` (maybe strip off the query with a regexp?), `url-unhex-string`, and `file-name-nondirectory`. That way you're never assuming anything about the host system's file separator character.

seblemaguer:
Hello,

I do agree. Thanks for the suggestions I wasn't aware of these functions. I am going to do that

seblemaguer:
I have added your suggestion and adapted mu4e functions to be more flexible. I still have to add asynchronous download

skeeto:
Whew! I ask you for some small changes and you go all out! I squashed your three commits into one, and made some tweaks and bug fixes along the way. So it's now merged as 5e290b5. As for what I fixed:

* In your PR there's a function call `(elfeed~get-default-download-dir)` (note the tilde). I assumed this was some sort of typo.

* `fpath` was being used as a free variable, which doesn't work correctly under lexical scope. The compiler warns about this: try running "make test" in the repository.

* Usage of `add-to-list`, which semantically makes no sense in lexical scope ... though the Elisp compiler has a hack in place specifically to support this. This function is really only a convenience for user configuration (e.g. .emacs, init.el) and shouldn't be used in normal Elisp code.

Except for the three new commands, I made all of the new functions "private" (via naming convention) and moved `elfeed-split-ranges-to-numbers` over into the general Elfeed library.

Thanks!

PR #92, opened 2015-06-03
adding version number

seblemaguer:
I'm currently finishing to prepare a sauron module for elfeed. I just need to have a version number in order to validate the load of the module. Do you think it's a good choice to add it  or did you have not add it on purpose ? If it's the first case, I propose to add it this way.

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -218,8 +218,7 @@ directory and saves all attachments in the chosen directory."
 
 (defun elfeed--download-enclosure (url path)
   "Download asynchronously the enclosure from URL to PATH."
-  (if (fboundp 'async-start)
-      ;; If async is available, don't hang emacs !
+  (if (require 'async nil t)
       (async-start
        (lambda ()
          (url-copy-file url path t))
--- a/elfeed.el
+++ b/elfeed.el
@@ -62,6 +62,10 @@
   "An Emacs web feed reader."
   :group 'comm)
 
+(defconst elfeed-version-string "1.1.2"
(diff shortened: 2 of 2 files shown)


skeeto:
There's no version string like this simply because it hasn't been
necessary so far. Is the way you did it here the canonical way to create
a package version string variable? Are there other examples of this in
other packages? As a counter-example, Magit has "magit-version" which
has a slightly different naming convention and doesn't list the version
directly in the source, instead deriving it at load time.

I'm hesitant to add it since there's already a version string listed in
the package headers, making this redundant. In theory package.el already
has the version string stored somewhere, though in the case of MELPA
this will be the MELPA timestamp version rather than the formal version.
If we do add a version variable, I want it to closely follow whatver
convention has already been established.

seblemaguer:
Hello,

thanks for the feedback. I have found an alternative solution which implies no modification in elfeed :) Sorry for the waste of time.

skeeto:
It's no waste of time! That may have proven to be important, I don't
know. I just wanted to make sure we did the most correct thing possible.

Could you share what your alternative solution is?

seblemaguer:
Ok so I'm reopening the issue as my other solution is not working properly.... For the version number, I checked so ERC, ejabber and twittering mode are actually using the same syntax.

skeeto:
Alright, so Magit, jabber.el, and twittering-mode all have *-version
variables. That's good enough for me. I renamed the variable in your
commit to elfeed-version, since the others don't have "-string" on the
end. Merged as 4b00344.

PR #100, opened 2015-09-04
silence byte-compiler

tarsius:


skeeto:
Thanks! I look forward to directly using the new unified clipboard
interface someday.

PR #101, opened 2015-09-16
minor correction to avoid invalid-function for async-start

seblemaguer:
As indicated on the title :D


skeeto:
Thanks, this is a good idea. I cherry-picked that last commit since you
have a couple other junk commits in the PR, then added with-no-warnings
to make the warning go away. The byte-compiler has a special check for
the previous fboundp pattern, so the with-no-warnings block wasn't
needed previously.

Merged as 3001cef.

PR #104, opened 2015-10-26
elfeed-search.el (elfeed-search-print): Display title with `C-h .`

daimrod:
Add a `kbd-help` property with the full title of the entry in the search view so
it can be displayed with `C-h .`.


skeeto:
I didn't know about this feature, thanks! Even after 10 years using
Emacs I'm still learning new things about it.

Issue #10, opened 2013-09-16
HTTP feeds fetched as HTTPS


cellscape:
Problem happens when elfeed fetches simultaneously HTTPS and HTTP feed. Emacs just stucks for a while and update doesn't proceed.

Here is part of backtrace:

    Debugger entered--Lisp error: (quit)
      re-search-forward("\\(^    Verify return code: .+\n---\n\\|^- Simple Client Mode:\n\\(\n\\|^\\*\\*\\* Starting TLS handshake\n\\)*\\)" nil t)
      open-tls-stream("feeds2.feedburner.com" #<buffer  *url-http-temp*-69304> "feeds2.feedburner.com" 80)
      network-stream-open-tls("feeds2.feedburner.com" #<buffer  *url-http-temp*-69304> "feeds2.feedburner.com" 80 (:type tls :nowait t))
      open-network-stream("feeds2.feedburner.com" #<buffer  *url-http-temp*-69304> "feeds2.feedburner.com" 80 :type tls :nowait t)
      ... 
      url-retrieve
      ... 
      elfeed--check-queue()
      ... 
      url-http-activate-callback()
      ...
      url-http
      url-https
      ...
      elfeed--check-queue()

Looking at backtrace, I can see that url-open-stream tries to use tls method for port 80 which is cleary invalid and results in my issue. Normally url-open-stream checks for url-gateway-method variable. However this variable is set by url-https down the stack and elfeed--check-queue asks to retreive new HTTP URL with url-gateway-method already set to tls.

One solution I could propose is to reset url-gateway-method to default value but I don't like this approach. Perhaps code that manages queue can be changed in some other way to fix the issue? 

BTW what do you think about using url-queue-retrieve?

skeeto:
I switched it over to `url-queue-retrieve` just now. No more custom queue framework. The only downside as far as I can tell is that it breaks differently in Windows than my original queue framework.

What prompted this is that YouTube just dropped their v2 API and the new v3 API is a whole lot more sensitive about rate-limiting, so Elfeed needs to be more careful about reusing connections to avoid trouble.

I'm going to close this since, if this bug is still around, it's formally an Emacs (url-queue) bug now.

PR #106, opened 2015-11-01
fix elfeed/search endpoint

d12frosted:
Not sure if it's right fix, but now `elfeed-search-filter` accepts `count`. And looks like passing `elfeed-web-limit` is the right thing. 

This PR is proposed in order to fix problem with `elfeed-web` when it gets internal server error on search.

### DIFF
--- a/web/elfeed-web.el
+++ b/web/elfeed-web.el
@@ -130,7 +130,7 @@
           (filter (elfeed-search-parse-filter q))
           (count elfeed-web-limit))
      (with-elfeed-db-visit (entry feed)
-       (when (elfeed-search-filter filter entry feed)
+       (when (elfeed-search-filter filter entry feed count)
          (setf (cdr tail) (list entry)
                tail (cdr tail))
          (when (< (cl-decf count) 0)


skeeto:
Thanks for bringing this up. I haven't really been maintaining
elfeed-web so I didn't know I broke it awhile back. It happened when I
introduced "limit" filters, which is # followed by a number (ex. #10 to
limit to 10 results).

Your change makes the error go away (most important!) but doesn't
implement the new filter. Turns out the limit filter is one that
elfeed-web was implementing internally to limit results. I made a
followup commit (be89df2) that builds on limit filter in elfeed-web,
simplifying it slightly.

d12frosted:
You're welcome. :smile: 

> Your change makes the error go away (most important!) but doesn't implement the new filter.

Ah, I feared that I am missing something. Honestly I didn't understand very well all parts of that function, I am pretty novice at emacs-lisp :blush: So thanks for covering my change! :smile: 

P. S. Thanks for the elfeed - you made my RSS reading simple and comfortable!

PR #125, opened 2015-12-07
Set `bookmark-make-record-function` buffer-locally

iqbalansari:
As per the documentation this should be set buffer-locally by modes


skeeto:
Thanks! Sorry if this broke something else in Emacs in the meantime.

iqbalansari:
No problem, that is a (completely understandable) trade-off of using development version instead of stable one. Thanks for Elfeed!

PR #126, opened 2015-12-07
Autoload bookmark-handler function

groks:
Now you can jump into elfeed from a saved bookmark without first having
to call `elfeed'.


skeeto:
Duh, I should have thought of this before. I never tested with it unloaded. Thanks! Merged as a8eaaae192645964e178bce3e6e3169d7112ebf6.

PR #127, opened 2015-12-07
Enable desktop support for restoring search buffer

groks:
I lost some data about which entries I had read when Emacs was not shut down cleanly (no problem with elfeed, or emacs). I already have desktop-mode enabled, and it already has the facility to save state when idle, so I added desktop support. Plus, of course, it restores you search buffer, which is nice.


skeeto:
Thanks! Merged as a985e7878d6edb7f4bb39c29860aa5c1eb51a6de.

Issue #34, opened 2014-06-30
code to create orgmode links


brabalan:
I have the following in my configuration files.

``` emacs-lisp
(defun elfeed-entry-as-html-link ()
  "Store an http link to an elfeed entry"
  (when (equal major-mode 'elfeed-show-mode)
    (let ((description (elfeed-entry-title elfeed-show-entry))
      (link (elfeed-entry-link elfeed-show-entry)))
     (org-store-link-props
      :type "http"
      :link link
      :description description))))

(add-hook 'org-store-link-functions 'elfeed-entry-as-html-link)
```

This way, when I org-capture in elfeed-show-mode using a template that includes a link, I get a nicely formatted link to the online linked entry. I use this to capture stuff I want to read later. Don't hesitate to use this if you want to integrate this small function to elfeed.

groks:
Here's another version:

    (defun my-elfeed-store-link ()
      "Store a link to an elfeed search or entry buffer."
      (cond ((derived-mode-p 'elfeed-search-mode)
             (org-store-link-props
              :type "elfeed"
              :link (format "elfeed:%s" elfeed-search-filter)
              :description elfeed-search-filter))
            ((derived-mode-p 'elfeed-show-mode)
             (org-store-link-props
              :type "elfeed"
              :link (format "elfeed:%s#%s"
                            (car (elfeed-entry-id elfeed-show-entry))
                            (cdr (elfeed-entry-id elfeed-show-entry)))
              :description (elfeed-entry-title elfeed-show-entry)))))

    (defun my-elfeed-open (filter-or-id)
      "Jump to an elfeed entry or search, depending on what FILTER-OR-ID looks like."
      (message "filter-or-id: %s" filter-or-id)
      (if (string-match "\\([^#]+\\)#\\(.+\\)" filter-or-id)
          (elfeed-show-entry (elfeed-db-get-entry (cons (match-string 1 filter-or-id)
                                                        (match-string 2 filter-or-id))))
        (switch-to-buffer (elfeed-search-buffer))
        (unless (eq major-mode 'elfeed-search-mode)
          (elfeed-search-mode))
        (elfeed-search-set-filter filter-or-id)))

    (org-add-link-type "elfeed" #'my-elfeed-open)
    (add-hook 'org-store-link-functions #'my-elfeed-store-link)

Rather than crate an *http:* link this creates a *elfeed:* link, which can link to either:

- an entry
- a search

For example, here's a search link:

[...]

PR #128, opened 2015-12-08
Support org-mode links to search and entry buffers

groks:
Also gathers meta-data for org-capture and exports links to HTML.

See #34 for more details.


skeeto:
As is, this would break [elfeed-org](https://github.com/remyhonig/elfeed-org) since it has the same name. The `provide` would clash. It's probably best to avoid the prefix "elfeed-org" since it's already been claimed by that package.

groks:
Good point. I've renamed elfeed-org- -> elfeed-link-

alphapapa:
Since #206 has been merged, this can be closed.

skeeto:
Closing via #206 per @alphapapa.

PR #142, opened 2015-12-23
Tolerate entries with non-compliant date formats.

miangraham:
Amazing package. Just tried elfeed for the first time a few days ago and I’m loving it.

I wanted to share and get thoughts on a quick change that made search results more usable for me.

A problem I had shortly after adopting elfeed was with entries containing non-compliant date formats—they would switch to the current time on each update and end up accumulating at the top of my search list, never moving down and out of the way.

To make this case a bit more tolerable, I changed the logic for non-parseable dates as follows:
* New entries will default to the current datetime (elfeed-float-time).
* Existing entries retain their existing date record.

Dates that parse successfully will still update the DB entry if the downloaded date string has changed to something new.

To make this fix easier I changed the behavior of elfeed-float-time (it now returns nil on failure) and added a test to suit. I don't see any negative side-effects from this on my end, but it could easily be done in another function if needed.

I could probably also write an end-to-end test to cover the high-level behavior change here, but wanted to see what you think first. Let me know if I’m missing anything.

Cheers and thanks for the great package!

### DIFF
--- a/elfeed-lib.el
+++ b/elfeed-lib.el
@@ -86,10 +86,19 @@ Examples: 2015-02-22, 2015-02, 20150222"
           (when (and (>= year 1900) (< year 2200))
             (float-time (encode-time 0 0 0 day month year t))))))))
 
+(defun elfeed-new-date-for-entry (entry new-date)
+  "Given an existing entry (nil if new) and a new date string,
+decide the entry's current date. Existing entries' dates are
+unchanged if the new date is not parseable. New entries with
+unparseable dates default to the current time."
+  (or (elfeed-float-time new-date)
+      (and entry (elfeed-entry-date entry))
+      (float-time)))
+
 (defun elfeed-float-time (&optional date)
   "Like `float-time' but accept anything reasonable for DATE,
-defaulting to the current time if DATE could not be parsed. Date
-is allowed to be relative to now (`elfeed-time-duration')."
+defaulting to nil if DATE could not be parsed. Date is allowed
+to be relative to now (`elfeed-time-duration')."
   (cl-typecase date
     (string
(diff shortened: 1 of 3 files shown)


skeeto:
Thanks, I wasn't aware of the problem of continually-changing dates
since I'm not subscribed to a feed with invalid dates (I bug authors to
fix it). If the date is completely missing, as it is for many feeds, it
will use the current date on keep it, but it's not so nice when it's
invalid, as you've seen.

Making elfeed-float-time return nil certainly makes more sense, since it
doesn't quietly hide the problem. That's definitely a good change.

There were a couple of problems with your patch, so I amended it to fix
them before merging.

* Your patch introduced a circular dependency. It's important that
  elfeed-lib.el doesn't use elfeed-entry-date from elfeed-db.el. It
  still works, but it's not ideal. You'll see the compiler complain
  about it ("make test"). I adjusted it to pull the date out before
  passing to elfeed-new-date-for-entry.

* If the date was missing altogether, the bad behavior you described
  would reappear. The new date would be nil, which would tell
  elfeed-float-time to return the current date (e.g. the optional date
  argument). I adjusted elfeed-float-time to return nil if given nil,
  and the argument is no longer optional. From a purity perspective,
  perhaps this should signal an error, but it plays nicer this way.

* The first line of the docstring should be a single, standlone
  sentence. I adjusted the docstrings to conform to this. Though I'm
  sure some of my older code violates this since I didn't know about it
  myself until a couple of years ago.

miangraham:
Oops, I'm new to package contributions and didn't know about the docstring convention at all. Also noted on the circular dependency.

Thanks for the adjustments--I'll give it a spin. Cheers!

PR #157, opened 2016-03-13
Enable yank to copy multiple feeds.

Fuco1:
I see no reason why `elfeed-search-yank` shouldn't support bulk copying, like tag/untag operations.  For example, when I get multiple feeds from youtube or other, I might want to run `youtube-dl` on all the links.  Right now, I use a bit of custom code to loop over the feeds and copy one by one.

This change enables me to just select the region, hit `y`, paste it to the terminal and be done with it.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -523,20 +523,21 @@ browser defined by `browse-url-generic-program'."
     (unless (use-region-p) (forward-line))))
 
 (defun elfeed-search-yank ()
-  "Copy the selected feed item to clipboard and kill-ring."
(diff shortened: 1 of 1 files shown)


Fuco1:
Notice also that the readme claims this already works, but it doesn't ;)

skeeto:
Thanks! The main reason Elfeed hadn't supported this before was that it didn't seem meaningful to me to put a list of delimited URLs on the clipboard. I've only ever cared about a single URL at a time (paste for wget, etc.). However, if someone has a good use for it, then there's no reason not to support it.

Fuco1:
Thanks, awesome!

PR #159, opened 2016-04-24
Support playing podcast using emms if emms available

seblemaguer:
I just added few functions to support playing directly enclosures if emms is installed

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -350,6 +350,52 @@ offer to save a range of enclosures."
       (elfeed-show-save-enclosure-multi)
     (elfeed-show-save-enclosure-single)))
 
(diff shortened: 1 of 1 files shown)


skeeto:
I like this. Did you mean for elfeed-show-play-enclosure-multi to only *save* the enclosures? That looks like it might be a copy-paste mistake. For now I just trimmed out the multi stuff and left it simply to playing a single enclosure at a time (KISS principle). Thanks!

seblemaguer:
I'm not quite sure that emms is a package included by default in emacs. That's why I put the when/require instruction. For the multi it was a copy/paste mistake but after more thinking I think it was useless too.

PR #162, opened 2016-05-19
Load index literally to avoid questioning when size gets to large

ksjogo:
My indexfile is 1.4M at the moment, resulting in the question to open literally or not. Setting this switch prohibits this question.

### DIFF
--- a/elfeed-db.el
+++ b/elfeed-db.el
@@ -285,7 +285,7 @@ This function increases the size of the structs in the database."
           (plist-put db :entries (make-hash-table :test 'equal))
           ;; Compiler will warn about this (bug#15327):
           (plist-put db :index (avl-tree-create #'elfeed-db-compare)))
-      (with-current-buffer (find-file-noselect index :nowarn)
+      (with-current-buffer (find-file-noselect index :nowarn t)
         (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


skeeto:
That's a very strange prompt. 1.4MB is small; small enough to be a human-edited text file (i.e. a large novel). I'm guessing this prompt is coming from Spacemacs: syl20bnr/spacemacs#3609.

If this is true, then you should instead adjust your `dotspacemacs-large-file-size`. This is similar to Emacs' own `large-file-warning-threshold`. The Spacemacs one defaults to a measly 1MB. IMHO, a more appropriate default is at least 100MB. Decoding a character encoding will generally be dwarfed by I/O, especially from spinning platters. It's nearly free when it's UTF-8 since that matches Emacs internal encoding. But maybe it's different on SSDs? The Emacs' info manual is 2.4MB on my system. Does it prompt you about that, too (`M-x info-emacs-manual`)?

Opening the index literally is definitely the wrong answer here. This would break all of Elfeed's Unicode support. It contains text, so we *want* character decoding.

ksjogo:
Oh, it was indeed the spacemacs handling.
Did not see any utf-8 artifacts, so did not notice anything there, just though it would disable modes which are unneeded for reading content in. Thanks for clearing that up.

PR #163, opened 2016-05-19
Synchronize prefix argument for browse-url in search and show

ksjogo:
C-u b now uses generic in view as in search

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -177,13 +177,17 @@ Defaults to `elfeed-kill-buffer'.")
   (elfeed)
   (elfeed-search-live-filter))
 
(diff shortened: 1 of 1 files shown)


skeeto:
I like this interface tweak. Is this a browse-url idiom from elsewhere in Emacs?

ksjogo:
Did not find (m)any reference inside my elpa dir/emacs source, so probably not that commonly used.
My use-case is mainly, that I have set generic to 'open' and non-generic to eww. For some feeds firefox is just better for reading and for some of my feeds eww is needed to get the actual post content and not only the excerpt. It could be interesting to add configurable default actions for each feed.

PR #166, opened 2016-06-16
Add missing defcustom keys

Fuco1:
Patch adds missing defcustom types/groups for curl-related settings.


skeeto:
Thanks! I totally forgot to go back and fill those in.

PR #172, opened 2016-08-28
Add endpoint for mark entry as read

areina:
Hi!
First, thanks for Elfeed and happy third birthday :-) 

I would like to show you a [mobile version of elfeed](http://github.com/areina/elfeed-cljsrn) that I'm currently developing. I'm open to any comment and I would really appreciate your feedback. 

For the navigation flow in the mobile app, I realized that it is necessary an option to mark an entry as read. I created a new endpoint for this, but I have some doubts that I would like to comment with you:
* This endpoint marks an entry as read searching by id. This is fine now, but probably in the future would be useful mark searching by tag or by feed. 
* I can't put the webid param in the url because it could contain a "/", so it should go as a query param.
* I'm not sure if `with-elfeed-db-visit` and `elfeed-db-return` is the efficient way to implement this. Would be nice if you can confirm it.

Thank you!

### DIFF
--- a/web/elfeed-web.el
+++ b/web/elfeed-web.el
@@ -150,6 +150,19 @@ advanced past it (long poll)."
         (push (httpd-discard-buffer) elfeed-web-waiting)
       (princ (json-encode update-time)))))
 
+(defservlet* elfeed/mark-read application/json (webid)
+  "Mark entry as read"
+  (with-elfeed-web
+   (let ((read-entry (with-elfeed-db-visit (e _)
+		       (when (string= webid (elfeed-web-make-webid e))
+			 (progn
+			   (elfeed-untag e 'unread)
+			   (elfeed-db-return e))))))
+     (if read-entry
+	 (princ (json-encode t))
+       (princ (json-encode '(:error 404)))
+       (httpd-send-header t "application/json" 404)))))
+
 (defservlet* elfeed/mark-all-read application/json ()
   "Marks all entries in the database as read (quick-and-dirty)."
   (with-elfeed-web


skeeto:
I'm excited to see you making use of this API! Your app looks slick, and this is the right way for it to interface with Elfeed. Thanks for doing this. I'll add a link to the README.

I've got some feedback for this change, including my own opinions on how to extend the API.

First, replace the manual webid lookup with a simple call to `elfeed-web-lookup`. That's what it's there for. For efficiency's sake, elfeed-web maintains the mapping of webids to entries.

Following RESTful practices, a GET should never manipulate the state of the database. A PUT is far more appropriate for changing tags. Unfortunately due to my short-sightedness years ago, simple-httpd, the web server behind elfeed-web, doesn't really distinguish the request types as part of routing. This could be checked manually or just ignored (for now). However, the structure of the request should more closely reflect a PUT.

I'd also like to make this endpoint more generalized to tagging. So rather than "mark-read" it should be something like "tags".

    PUT /elfeed/tags/<webid>

The naive approach would be to simply provide the exact new list of tags. That is, you would fetch the current set of tags, manipulate the set, then send it back. A much smarter approach is more along the lines of "add tag" and "remove tag", which, unlike the naive solution, is atomic.

Going with the latter, I'm picturing this. The client sends a JSON object with optional "add" and "remove" properties, each refering to a list of tags.

    PUT /elfeed/tags/<webid>
    {"add": ["foo", "bar"], "remove": ["unread"]}

This also leaves the potential option of mass tag changes across multiple webid objects ("mark all as read") in a single request, should you think you need it.

    PUT /elfeed/tags
    {"entries": [<webid>, <webid>], "remove": ["unread"]}

What do you think?

areina:
I totally agree about PUT requests. 

About the endpoint for tagging,  I'm not sure if it is correct in terms of RESTful practices. In
```
PUT /elfeed/tags/<webid>
```
I would understand that you are updating a tag resource. if you want to update the tags of an entry,  I think the proper endpoint would be:
```
PUT /elfeed/entries/<webid>/tags
```
BTW,  I don't know if makes sense to build a complete restful api for elfeed, so maybe having a single endpoint for tagging as your last example could be a good solution.

skeeto:
I think you're right about /elfeed/entries/<webid>/tags. But as you
point out, Elfeed probably doesn't need a full API. Additionally, tags
and elfeed-meta are the only two ways entries are intended to be mutated
by the user anyway. Finally, my dumb defservlet macro doesn't support
that fancy format anyway.

In a branch I pushed ac91aa5 which implements the simpler version, just
the "add" and "remove" fields. It only accepts PUT requests, and it
returns the current set of tags on success. Does this look reasonable
for your needs?

For CSRF resistance, on master I pushed 7d88bcd, which makes webids
difficult to predict.

areina:
Thank you!

It looks well for what I need, but I think your last suggestion was more interesting:

```
PUT /elfeed/tags
{"entries": [<webid>, <webid>], "remove": ["unread"]}
```
I think this endpoint is more correct in terms of RESTful good practices, and it works well for single or multiple entries. Do you think would be useful this implementation? If not, I can go ahead and update my code to use https://github.com/skeeto/elfeed/commit/ac91aa58948d27d3e3964fdbd2c1e70f49b4af7e

skeeto:
Ok, I abandonded the webid part of the entrypoint and just coded it as
described with the "entries" property. It's now on master.

areina:
Great!  Thank you so much.  Closing this PR

Issue #181, opened 2016-10-20
elfeed-add-feed, x-get-selection-value, and more


holomorph:
hi!

`x-get-selection-value` signals an error, complaining "No selection is available" here.  This is unfortunate because it then prevents falling through to the kill ring (which is where my feed is!) Perhaps you have a particular reason for using this, but for a similar functionality in transmission.el I used `x-get-selection`, which just returns nil if no selection data is found.

Oh, also, supposedly using the `INITIAL-CONTENTS` argument of read-from-minibuffer is deprecated.  I vaguely remember reading about that recently; IIRC the elisp manual has an explanation: ah yes, from the `read-from-minibuffer` index item, you get linked to `(info "(elisp) Initial Input")`

I also did it slightly differently (with a hook) and now I'm thinking it might be slightly more useful to instead accumulate the results from each thing-that-might-give-you-a-string, and put them in the `DEFAULT-VALUE` argument of `read-from-minibuffer`, so one can M-n through the different possible sources.

holomorph:
oh, you might also consider adding some things to look at point; e.g. `url-get-url-at-point`, `eww-links-at-point`, `(get-text-property (point) 'shr-url)`, etc.

sorry for putting half a dozen topics into the issue :)

skeeto:
I can't think of any reason why I used `x-get-selection-value` over `x-get-selection`. I probably just saw it first. Someday when I drop support for Emacs 24, this will be replaced with the Emacs 25 unified clipboard functions.

As for `INITIAL-CONTENTS`, the manual says it's "*mostly* deprecated." :-) I don't normally use it, but I like how it works in this particular case. I also like your idea for collecting the various options and presenting them all. Since I never actually use `elfeed-add-feed` myself outside of testing, implementing this isn't a priority for me, but I'd accept a well-done PR for it.

PR #184, opened 2016-12-16
Collect feeds from multiple sources for elfeed-add-feed (#181)

holomorph:
finally got around to it!

### DIFF
--- a/elfeed-lib.el
+++ b/elfeed-lib.el
@@ -13,6 +13,7 @@
 (require 'cl-lib)
 (require 'time-date)
 (require 'url-parse)
+(require 'url-util)
 
 (defun elfeed-expose (function &rest args)
   "Return an interactive version of FUNCTION, 'exposing' it to the user."
@@ -241,6 +242,18 @@ systems."
        (and (fboundp 'w32-get-clipboard-data)
             (funcall 'w32-get-clipboard-data))
        (current-kill 0 :non-destructively))))
+
+(defun elfeed-get-link-at-point ()
+  "Try to a link at point and return its URL."
+  (or (get-text-property (point) 'shr-url)
+      (and (fboundp 'eww-current-url)
+           (funcall 'eww-current-url))
+      (get-text-property (point) :nt-link)))
+
+(defun elfeed-get-url-at-point ()
+  "Try to get a plain URL at point."
+  (or (url-get-url-at-point)
+      (thing-at-point 'url)))
 
 (defun elfeed-move-to-first-empty-line ()
   "Place point after first blank line, for use with `url-retrieve'.
--- a/elfeed.el
+++ b/elfeed.el
@@ -131,6 +131,18 @@ when they are first discovered."
   :type '(repeat (choice string
                          (cons string (repeat symbol)))))
 
+(defcustom elfeed-feed-functions
+  '(elfeed-clipboard-get
+    elfeed-get-link-at-point
+    elfeed-get-url-at-point)
+  "List of functions to use to get possible feeds for `elfeed-add-feed'.
+Each function should accept no arguments, and return a string or nil."
+  :group 'elfeed
+  :type 'hook
+  :options '(elfeed-clipboard-get
+             elfeed-get-link-at-point
+             elfeed-get-url-at-point))
+
 (defcustom elfeed-use-curl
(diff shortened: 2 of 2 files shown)


skeeto:
Thanks! I like the way you've implemented it, especially with the `elfeed-feed-functions` customization, except that it's not applied correctly to `elfeed-add-feed`. You use the result of `elfeed-candidate-feeds` (a list) as the `DEFAULT-VALUE` argument, which the documentation says "if non-nil, should be a string." It's definitely not working correctly because of this, though it might appear to be because it gets a generic minibuffer history.

I suggest using the `hist` argument instead. Unfortunately due to the archaic interface, it requires a dynamically scoped variable. Below I declare a special variable `elfeed--add-hist`, which gets bound around the `read-from-minibuffer` call.

~~~el
(defvar elfeed--add-hist)

(defun elfeed-add-feed (url)
  "Manually add a feed to the database."
  (interactive
   (list
    (let ((elfeed--add-hist (elfeed-candidate-feeds)))
      (read-from-minibuffer "URL: " nil nil nil 'elfeed--add-hist))))
  ...)
~~~

Beyond this there is still another issue. The default is just an empty string (`""`), which the rest of the function does not handle gracefully. Perhaps instead it should use the `DEFAULT-VALUE` argument, with the prompt being dynamically created to indicate what the default value is. This should gracefully handle the case when there are no candidates.

Also, I suggest using `cl-pushnew` in `elfeed-candidate-feeds` rather than plain `push`, in case the same URL is returned by multiple sources.

~~~el
(when (not (zerop (length val)))
  (cl-pushnew val res :test #'equal))
~~~

holomorph:
My `read-from-minibuffer` reads
```
Sixth arg DEFAULT-VALUE, if non-nil, should be a string, which is used
  as the default to `read' if READ is non-nil and the user enters
  empty input.  But if READ is nil, this function does _not_ return                        
  DEFAULT-VALUE for empty input!  Instead, it returns the empty string.

  Whatever the value of READ, DEFAULT-VALUE is made available via the
  minibuffer history commands.  DEFAULT-VALUE can also be a list of
  strings, in which case all the strings are available in the history,
  and the first string is the default to `read' if READ is non-nil.
```
I harbor much dislike for the `HIST` argument :)  Since `minibuffer-history` is already dynamic I just bound that when trying to work around this when turning on `lexical-binding` in `replace.el`.  Also, in this case I don't think `HIST` makes as much sense, since here we are trying to provide defaults which we seek with `M-n`, not history we revisit with `M-p`.  Having written this, the above docstring now appears confusing since it refers to both defaults and actual minibuffer history as "history"

Perhaps some logic indicating the default as the car of DEFAULT-VALUE in the prompt, and changing the result to said default if we get an empty string would be alright?
``` elisp
(defun elfeed-add-feed (url)
  "Manually add a feed to the database."
  (interactive
   (list
    (let* ((feeds (elfeed-candidate-feeds))
           (prompt (if feeds (concat "URL (default " (car feeds)  ": ")
                     "URL: "))
           (res (read-from-minibuffer prompt nil nil nil nil feeds)))
      (if (and (zerop (length (elfeed-cleanup res))) feeds)
          (car feeds)
        (user-error "????")))))
  ...)
```
Good call on using `cl-pushnew`, thanks.

skeeto:
I understand now. That's what I get for not reading into the next
paragraph. I noticed M-p was a generic history and thought it wasn't
working, but I now see the correct items with M-n. What you're doing
with DEFAULT-VALUE is indeed fine.

Your change to the prompt is exactly what I was looking for, too. Except
that it needs to return res if it's not empty. And, of course, something
more informative than "????" as an error message.

Make the prompt change and the cl-pushnew change, and I'll merge it.

holomorph:
I did a couple other things that seemed sensible to me:
- add `elfeed-looks-like-url-p` as a condition to add a URL to the list of candidates
- Move `elfeed-clipboard-get` to the end of the hook.  I figured the cases where point happened to be in EWW or on top of a URL somewhere, if someone is doing M-x elfeed-add-feed there then that is probably the link they are after.

skeeto:
Squashed and merged as 455e58e. Thanks!

holomorph:
Nice, thanks

PR #200, opened 2017-03-04
Add elfeed-show-entry-author.

dellison:
Some of the feeds that I subscribe to have a few contributing authors, and if this is available in the metadata for an entry, I think it's nice to have an option to show it.

I added this as a customizable option which is off by default. The "author" is taken from the metadata of `elfeed-show-entry`, and is only shown in the `elfeed-entry` buffer if it's present and the option is enabled (i.e. `elfeed-show-entry-author` is non-nil).

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -18,6 +18,11 @@
   :group 'elfeed
   :type 'boolean)
 
+(defcustom elfeed-show-entry-author nil
+  "When non-nil, show the entry's author (if it's in the entry's metadata)."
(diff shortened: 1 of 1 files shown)


skeeto:
Great patch, thanks! Did you disable it by default in order to make this
change less intrusive? I think this would be reasonable to have enabled
by default. Was there another reason?

dellison:
Thanks! And yes, you're exactly right, I left if off just to have it be less intrusive. Enabling it by default would be fine with me. :)

PR #210, opened 2017-04-18
Allow user apply autotags to all entries with the latest rules easily

fasheng:
I'm new to elfeed and really like it, thanks for the great work!

At first, I import my feed list to elfeed and it works, then realize elfeed allow user add autotags for each feed, it really cool! But the old entries just keep non-tags as before, and I don't want to delete the database so wrote a patch to do such work. Wish it helps for others~

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -576,6 +576,18 @@ Only a list of strings will be returned."
     (customize-save-variable 'elfeed-feeds elfeed-feeds))
   (elfeed-update-feed url))
 
+(defun elfeed-apply-autotags-all ()
+  "Applay autotags to all entries with the latest rules."
+  (interactive)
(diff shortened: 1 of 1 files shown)


skeeto:
I didn't accept your particular implementation, but I did add
elfeed-apply-autotags-all in ea70279. It does the same thing, that name
parallels the name elfeed-apply-hooks-now, and my version is written
more correctly. Thanks for sparking the idea, though!

Just to provide some feedback: The proper way to iterate over the
entries is with with-elfeed-db-visit. The hash table isn't necessarily a
stable interface, though I admit the name doesn't indicate such. The
macro also supplies the entry and feed at the same time, making the body
a trivial one-liner.

fasheng:
Cool, let me benefit from it, simple and clean, it works here! :)

PR #213, opened 2017-04-29
Make curl support custom HTTP request method and data so let ownCloud News work with elfeed possible

fasheng:
This is a complex commit, include:
- support ownCloud News as backend

- improve curl interface to support HTTP PUT request method

- add function elfeed-is-status-error to avoid repeat code

- adjust elfeed-search--header-1 so that search view header will be
  updated once curl fetch operation started to avoid update header only
  after ownCloud News fetched

- adjust elfeed-tag/elfeed-untag so that once unread/star tag modified
  the state of ownCloud News will be synced

- fill two test cases for parsing ownCloud News feeds and entries

This commit will not break the original functions, user could switch value
between local and ocnews for variable elfeed-feed-source for therir
data structures are compatibility.

Try to experience this feature with following steps:
1.  fetch nextcloud image and run it(recommend nextcloud instead of owncloud)
    
        docker pull nextcloud
        docker run --rm -p 80:80 nextcloud
2.  open <http://127.0.0.1> in browser, setup the nextCloud
    1.  create admin user and select database to SQLite, then press "Finish setup"
    2.  press left top popup menu and select "+Apps", select
        "Multimedia", and enable "News" app
    3.  press left top popup menu and switch to "News" app, then
        subscribe some feeds, that's all
3.  setup elfeed, then switch to search view and and press G to update entries.
    
        (setq elfeed-feed-source 'ocnews)
        (setq elfeed-ocnews-url "http://127.0.0.1")
        (setq elfeed-ocnews-username "user")
        (setq elfeed-ocnews-password "password")

**BTW:** Really like selfhost rss server such like ownCloud News, and experienced for several days, just works fine, so wish this PR could be accept ;)

### DIFF
--- a/elfeed-curl.el
+++ b/elfeed-curl.el
@@ -267,7 +267,7 @@ Use `elfeed-curl--narrow' to select a header."
            do (setf location (elfeed-update-location location value))
            finally return location))
 
-(defun elfeed-curl--args (url token &optional headers)
+(defun elfeed-curl--args (url token &optional headers method data)
   "Build an argument list for curl for URL.
 URL can be a string or a list of URL strings."
   (let ((args ()))
@@ -282,6 +282,8 @@ URL can be a string or a list of URL strings."
     (dolist (header headers)
       (cl-destructuring-bind (key . value) header
         (push (format "-H%s: %s" key value) args)))
+    (when method (push (format "-X%s" method) args))
+    (when data (push (format "-d%s" data) args))
     (setf args (nconc (reverse elfeed-curl-extra-arguments) args))
     (if (listp url)
         (nconc (nreverse args) url)
@@ -297,12 +299,12 @@ URL can be a string or a list of URL strings."
   (elfeed-curl--decode)
   (current-buffer))
 
-(defun elfeed-curl-retrieve-synchronously (url &optional headers)
+(defun elfeed-curl-retrieve-synchronously (url &optional headers method data)
   "Retrieve the contents for URL and return a new buffer with them.
 HEADERS is an alist of additional headers to add to the HTTP request."
   (with-current-buffer (generate-new-buffer " *curl*")
     (setf elfeed-curl--token (elfeed-curl--token))
-    (let ((args (elfeed-curl--args url elfeed-curl--token headers))
+    (let ((args (elfeed-curl--args url elfeed-curl--token headers method data))
           (coding-system-for-read 'binary))
       (apply #'call-process elfeed-curl-program-name nil t nil args))
     (elfeed-curl--parse-write-out)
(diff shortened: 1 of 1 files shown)


elfeed-curl.el
```diff
@@ -392,34 +394,36 @@ results will not."
...
           (push key keys)
           (push entry (gethash key table nil)))))
     (dolist (key (nreverse keys))
       (let ((entry (gethash key table)))
         (when entry
           (let ((rotated (list (nreverse (cl-mapcar #'car entry))
                                (nreverse (cl-mapcar #'cadr entry))
-                               (cl-caddar entry))))
+                               (cl-caddar entry)
+                               (elt (car entry) 3)
```
fasheng: I'm not sure, looks the code style is a little difference, but works..

fasheng:
Fixed a sync unread/starred repeatly issue just now.

But there is still one that the top header message sometimes keep "0 feeds pending, 1 in process..." shown even the ownCloud update operation finished :/

 Any advice, please~

fasheng:
Update:
- Fix CPU hang up issue when downloading large JSON file, for example
  more than 10000 entries
- Works both with curl and url-retrieve now (`url-queue-retrieve` do not
  support url-request-method/url-request-data under let expression, so
  use `url-retrieve` instead. Yes, a bug is that search header will not show
  updating information for url-retrieve)
- Split commits

fasheng:
Hi @skeeto, excuse me again.

I just like elfeed and want it support ownCloud News and other self hosting RSS readers, but perhaps it's not a goal of elfeed. So I plan to split the code to a separate package, but it will be impossible if elfeed  don't support curl HTTP **PUT** method. I will drop the ownCloud related code and left curl one, hope it could be accepted and any advice is welcome :)

Have a good day!

skeeto:
Sorry about the radio silence. This is a big change that adds a feature 
that doesn't personally interest me — I never heard of ownCloud before 
this and I don't use it — so I kept putting off reviewing your PR. 
Though I am perfectly happy to make the hooks/changes you need to make 
your alternative backend work. Maybe this will open the door to 
pluggable alternative database backends.

What do you think about turning your new "method" and "data" arguments 
into keyword arguments via cl-defun? The "headers" argument would be 
converted with this change. There's technically the chance this might 
break someone's config somewhere, but I'd rather do it right from the 
get-go. There's not really any reason for these arguments to be 
positional, and "headers" is only positional because I didn't anticipate 
adding more arguments.

fasheng:
Thanks, I saw the last code, that's what people want, it's more clearly. :)

elfeed works great, but sometimes people want a self host server because they want to read RSS reader in any platforms, and the articles and unread/star tags could sync each other. elfeed is great and powerful enough so it will be cool to support such cloud RSS readers. The following work will be started soon, thanks for the help and welcome any advice 😄

skeeto:
Do you have your original ownCloud commits somewhere? I can't find them 
now that you've backed them off from the PR.

fasheng:
Oh, sorry not promptly notified. I just start a separate package named `elfeed-backends`. It works but full of def-advices. So I plan to submit another PR to support some hooks in elfeed to avoid such advices, the work will be finished soon.

https://github.com/fasheng/elfeed-backends

fasheng:
Though I prefer full code could be merged into elfeed, once the work finished, you can determine it, sir ~

fasheng:
The first step `elfeed-backends` complete now, I'm afraid you not very interesting so now the goal is to act as a extension for `elfeed`, just like `elfeed-org`. I will submit a new PR to help do this, let's discuss there :)

PR #214, opened 2017-04-29
Use run-mode-hooks in show- and search-mode.

atheriel:
Since Elfeed rolls its own major modes instead of using `define-derived-mode`, it runs its own hooks manually using `run-hooks`.

However, `define-derived-mode` uses `run-mode-hook`, not `run-hooks`. As a consequence,  Elfeed's major modes do not currently run the universal `change-major-mode-after-body-hook` and `after-change-major-mode-hook`, which some packages rely on to detect major mode changes.

This PR replaces `run-hooks` with `run-mode-hooks` in `elfeed-search-mode` and `elfeed-show-mode` to resolve this issue.

I don't know what the original rationale was for using the regular `run-hooks`, so please feel free to reject this if there is a good reason!

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -192,7 +192,7 @@ When live editing the filter, it is bound to :live.")
   (add-hook 'elfeed-update-init-hooks #'elfeed-search-update--force)
   (add-hook 'kill-buffer-hook #'elfeed-db-save t t)
   (elfeed-search-update :force)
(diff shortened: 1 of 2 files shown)


skeeto:
Merged. The reason for run-hooks instead of run-mode-hooks is simply
because I didn't know any better. Thanks!

atheriel:
Cheers!

PR #221, opened 2017-06-01
Add elfeed-update-all

fasheng:
Update all feed in a special function so [elfeed-sources](https://github.com/fasheng/elfeed-sources) could make advice to it instead of `elfeed-update`.

I'm not sure if the function name is well, you could adjust it if necessary, sir ;)


skeeto:
I don't like having both an elfeed-update and elfeed-update-all. It's 
redundant and confusing. So far I still think hooking into this via a 
protocol registry is more appropriate.

fasheng:
Yes, so I ask you to give new names ;)
How about just separate the methods and just let `elfeed-sources` works, firstly.

PR #236, opened 2017-09-02
Use hard-coded list instead of dynamic discovery

Fuco1:
Because of the internal changes in Emacs cl implementation the "introspection" I've used to get the list of properties of the structure stopped working in Emacs 25.2+

I think it's not worth the effort to support multiple Emacs versions/cl implementations so I've just replaced the reflection with a static list.  Given how stable the elfeed data structures are there should never be a need to change the code very much.


skeeto:
You are correct about data structure stability. Thanks to the meta plist 
slot, I do not expect to ever change any of these slots ever again. New 
stuff will be tacked onto the plist. Thanks!

PR #242, opened 2017-10-01
Allow passing a flag to ‘elfeed-add-feed’ to save the new feed list.

joewreschnig:
The flag is always set when using the command interactively.

Also use ‘equal’ as the test predicate for ‘cl-pushnew’ as the list
contains strings.

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -584,8 +584,10 @@ Only a list of strings will be returned."
        nil))
     (nreverse res)))
 
-(defun elfeed-add-feed (url)
-  "Manually add a feed to the database."
+(defun elfeed-add-feed (url &optional save)
+  "Manually add a feed to the database.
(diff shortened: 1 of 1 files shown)


skeeto:
Thanks! It's good to have more precise control over side effects.

If I dig into the function's history I can see how the cl-pushnew went 
wrong. It originally used #'string= for comparisons. When I added 
autotagging in 2014, this was neglected and broken for awhile. I "fixed" 
it by allowing it default to #'eql. Doh!

I made one follow-on commit to convert your new positional argument into 
a keyword argument. If this function accumulates more options in the 
future — particularly boolean options like this — it's much nicer when 
these are each keyword arguments. One only needs to look at 
completing-read to see organic growth of positional arguments going 
wrong.

Issue #216, opened 2017-05-04
Tips: toggle search filter for current entry feed


fasheng:
Here is a rough implementation which toggle search string with the feed title for elfeed search filter looks not support feed url now. Even though, it works most time and really helpful for me.

```emacs-lisp
(defun my-elfeed-search-toggle-feed-filter ()
  "Toggle filter with current feed title."
  (interactive)
  (let* ((entry (elfeed-search-selected :ignore-region))
         (feed (when entry (elfeed-entry-feed entry)))
         (feed-title (when feed (elfeed-meta feed :title))))
    (when feed-title
      (if (string-match-p feed-title elfeed-search-filter)
          (progn
            (setq elfeed-search-filter
                  (replace-regexp-in-string " +" " "
                                            (replace-regexp-in-string
                                             (regexp-quote feed-title) ""
                                             elfeed-search-filter)))
            (elfeed-search-update--force))
        (progn
            (setq elfeed-search-filter (concat elfeed-search-filter
                                               " " feed-title))
            (elfeed-search-update--force))))))
(define-key elfeed-search-mode-map (kbd "<") 'my-elfeed-search-toggle-feed-filter)
```

PR #244, opened 2017-10-30
Add keybinding < to toggle current feed url in filter (#216)

fasheng:
Glade to see new `=url` syntax support in filter, and this code provide keybinding `<` to toggle current feed url in filter with it, just like org-mode agenda view did :)


minad:
Added feed filter toggle in 6902d59c76664a6ca99d94b787a777551939ce91

PR #245, opened 2017-10-30
Support extra protocols in elfeed-feeds

fasheng:
<del>The extra protocol type will be prefixed in the url, for example `owncloud+https://user:pass@myhost.com`, and the related update function should be named as `elfeed-protocol-owncloud-update`</del>.

The extra protocol means special protocol besides http and https in feed urls, which will be prefixed and split by "+" in the url, for example "owncloud+https://user:pass@myhost.com". User could register its update function with `elfeed-protocol-register`.

As we disscussed before(#220 #221), the last code follows your advice to define special protocol urls in `elfeed-feeds' to make ownCloud or other RSS service could works. And I rename my project to [elfeed-protocol](https://github.com/fasheng/elfeed-protocol), too.

Now we could use even multiple ownCloud News together with normal feed urls in `elfeed-feeds`, here is a demo code:
```emacs-lisp
(setq elfeed-feeds (list
                    "https://example.com/rss"
                    "owncloud+https://user1:pass1@myhost.com"
                    (list "owncloud+https://user2@myhost.com"
                          :password "password/with|special@characters:"
                          :autotags '(("example.com" comic))
                          )))
```

The only problem is it will break the origin autotags rules for protocol feeds, so there is workaround code to make such `elfeed-org` work with `elfeed-protocol`, but in my option it's a little strange XD:
```emacs-lisp
(defadvice elfeed (after configure-elfeed-feeds activate)
  "Make elfeed-org autotags rules works with elfeed-protocol."
  (setq elfeed-protocol-tags elfeed-feeds)
  (setq elfeed-feeds (list
                    (list "owncloud+https://user@myhost.com"
                          :password "password/with|special@characters:"
                          :autotags elfeed-protocol-tags
                          ))))
```

BTW: Sorry for the late code, it's so busy in recent months, and I will finish the project in few days. We all know that WIP long time is not a good habit ;)

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -173,6 +173,19 @@ Each function should accept no arguments, and return a string or nil."
              elfeed-get-url-at-point
              elfeed-clipboard-get))
 
+(defcustom elfeed-protocols ()
+  "List of all registered extra protocols in Elfeed.
+
+Could be modified by `elfeed-protocol-register' and
+`elfeed-protocol-unregister'.
+
+For example,
+  (elfeed-protocol-register \"owncloud\" 'elfeed-protocol-owncloud-update)
+  (elfeed-protocol-unregister \"owncloud\")
+"
+  :group 'elfeed
+  :type '(repeat (cons string symbol)))
+
 (defcustom elfeed-use-curl
   (not (null (executable-find elfeed-curl-program-name)))
   "If non-nil, fetch feeds using curl instead of `url-retrieve'."
@@ -500,7 +513,9 @@ Only a list of strings will be returned."
   (dolist (feed elfeed-feeds)
     (unless (cl-typecase feed
               (list (and (stringp (car feed))
-                         (cl-every #'symbolp (cdr feed))))
+                         (if (not (elfeed-protocol-feed-p (car feed)))
+                             (cl-every #'symbolp (cdr feed))
+                           t)))
               (string t))
       (error "elfeed-feeds malformed, bad entry: %S" feed)))
   (cl-loop for feed in elfeed-feeds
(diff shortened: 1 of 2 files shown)


elfeed.el
```diff
@@ -513,7 +528,8 @@ Only a list of strings will be returned."
                  (or (elfeed-feed-url url-or-feed)
                      (elfeed-feed-id url-or-feed))
                url-or-feed)))
-    (mapcar #'elfeed-keyword->symbol (cdr (assoc url elfeed-feeds)))))
+    (unless (elfeed-protocol-feed-p url)
+      (mapcar #'elfeed-keyword->symbol (cdr (assoc url elfeed-feeds))))))
```
fasheng: Make properties in protocol url like `("proto+https://user2@myhost.com" :prop1 "xxx" :prop2 "xxx")` works

elfeed.el
```diff
@@ -538,38 +554,47 @@ Only a list of strings will be returned."
...
+  (if (elfeed-protocol-feed-p url)
+      (let* ((proto-type (elfeed-protocol-type url))
+             (update-func (elfeed-protocol-update-func proto-type)))
+        (if update-func
+            (progn
+              (funcall update-func (elfeed-protocol-url url))
+              (run-hook-with-args 'elfeed-update-hooks url))
+          (elfeed-log 'error "There is not updater for protocol %s"
+                      proto-type)))
+    (elfeed-with-fetch url
```
fasheng: Check for a protocol updater and the rest code just not changed

fasheng:
Sorry to trouble again, sir. This should be the final commit and it didn't change too much, so hope to take several minutes to review the code and any advice is welcome.

And my original project will be split with different protocol names, here is [elfeed-owncloud](https://github.com/fasheng/elfeed-owncloud). If possible, such like `elfeed-ttrss` `elfeed-newsblur` will be implemented in future, too.

Have a good day!

fasheng:
This patch is not necessary now, anyone want ownCloud News, Tiny Tiny RSS or other self-hosting RSS readers works with elfeed just vist [here](https://github.com/fasheng/elfeed-protocol), and thanks to skeeto's great work for elfeed, too. Enjoy!

PR #251, opened 2017-11-16
Fix: (elfeed-search-set-filter) Treat empty string as nil

alphapapa:
The docstring says that, when the new filter is nil, it will be reset to the default value, but when called interactively, if the user entered an empty filter string, the empty string prevented the default filter from being restored.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -566,7 +566,9 @@ expression, matching against entry link, title, and feed title."
             (if current-prefix-arg "" elfeed-search-filter)))))
   (with-current-buffer (elfeed-search-buffer)
     (setf elfeed-search-filter
-          (or new-filter (default-value 'elfeed-search-filter)))
+          (or (unless (string-empty-p new-filter)
+                new-filter)
+              (default-value 'elfeed-search-filter)))
     (elfeed-search-update :force)))
 
 (defun elfeed-search--update-list ()


skeeto:
It's true that there's no way to pass nil as the argument interactively. 
However, an empty string is still a valid filter: "show me everything." 
After this change, the empty filter could still be achieved with a 
whitespace string (" "), but that's awkward and not obvious. The empty 
string and nil are distinct values, even in this context.

Personally when I want to reset the search filter I just kill the search 
buffer. It's a buffer local variable, so its value is discarded with the 
buffer. If that's inconvenient, it's trivial for anyone to make their 
own binding that actually does call elfeed-search-set-filter with nil. 
In fact, it's pretty common to bind specific keys to specific filter 
strings (i.e. "a" sets the filter to "@1-month +audio").

alphapapa:
Good point, I didn't realize that an empty string means to show everything.

A couple of ideas:

* Add a command to reset to default filter.
* Make `elfeed-search-live-filter` reset to the default filter when called with a universal argument.

If you are interested in either of those, I can submit a PR for them.  Otherwise I'll just add a command in my own config.

Thanks.

skeeto:
The second idea sounds useful, but it's at odds with the universal 
argument for the non-live filter. I'd prefer to mirror that behavior 
since it would be less surprising.

I guess I'm not opposed to adding a function to reset to the default 
filter. I've actually already done this in my own configuration, and 
bound it to "h" (like "home" in a web browser). If you want to do a PR 
for that, I'll accept it.

alphapapa:
Well, I ended up making a hydra, which I'm satisfied with:
```el
(cl-defun ap/elfeed-view-tag (str &key add toggle)
    "Concat STR with @6-months-ago +unread.
If ADD is non-nil, return current filter with STR added.  If
TOGGLE is non-nil, return current filter with STR added or
removed from it."
    (cond (add (concat elfeed-search-filter " " str))
          (toggle (if (string-match (regexp-quote str) elfeed-search-filter)
                      (s-replace-regexp (rx-to-string `(seq (optional (1+ space)) ,str (optional (1+ space))))
                                        "" elfeed-search-filter 'fixedcase 'literal)
                    (concat elfeed-search-filter " " str)))
          (t (concat "@6-months-ago +unread " str))))

  (defhydra ap/elfeed-search-view (elfeed-search-mode-map "v" :color blue)
    "Set elfeed-search filter tags."
    ("d" (elfeed-search-set-filter nil) "Default")
    ("n" (elfeed-search-set-filter (ap/elfeed-view-tag "+news")) "news")
    ("p" (elfeed-search-set-filter (ap/elfeed-view-tag "+politics")) "politics")
    ("s" (elfeed-search-set-filter (ap/elfeed-view-tag "-starred" :toggle t)) "unstarred")
    ("t" (elfeed-search-set-filter (ap/elfeed-view-tag "+tech")) "tech"))
```
So I just `vd`  to get back to default view.  Good enough for me.  :)  Thanks.

PR #263, opened 2018-01-23
Shut up the byte compiler for org 9.0+ users

jabranham:


skeeto:
Thanks! I'd noticed this when testing with the bleeding edge version of 
Emacs, but hadn't paid much attention yet.

jabranham:
Yeah, the upcoming Emacs 26.1 release has org 9.1.6, which complains when you use org-add-link-type.

Thanks!

Issue #208, opened 2017-04-12
Symbol's function definition is void: elfeed-link-store-link


ReichKV:
After upgrading to new elfeed I start emacs and get following mistake:

Warning (initialization): An error occurred while loading ‘/home/r/.emacs.d/init.el’:  Symbol's function definition is void: elfeed-link-store-link .. .... . .

(if i comment (use-package elfeed) i still get this error. I dont see the error only if remove elfeed.)

How to solve this problem?

skeeto:
Ah, crap, sorry. This has to do with merging #206 this morning. The
problem is that the new autoload doesn't actually load elfeed-link. You
can work around this manually for the moment by temporarily adding this
to your init before either Elfeed or Org is loaded:

    (require' elfeed-link)

I need to figure out how to load elfeed-link without firing the
eval-after-load twice.

ReichKV:
Thank you, the new update works.

skeeto:
Thanks for following up!

PR #268, opened 2018-02-19
Autoload all of elfeed-link

tummychow:
Due to the fix for #208, elfeed gets loaded after org-mode. I use org-mode extensively but do not use elfeed-link at all, and I don't want elfeed to be loaded in those situations. This provides an alternative solution to #208 that does not force elfeed to be loaded after org-mode. Instead, it binds autoloads to the functions that org-mode would invoke. I tested an elfeed link to confirm that elfeed would actually get loaded if you attempted to open the link.

### DIFF
--- a/elfeed-link.el
+++ b/elfeed-link.el
@@ -16,6 +16,7 @@
 (require 'elfeed-show)
 (require 'elfeed-search)
 
+;;;###autoload
 (defun elfeed-link-store-link ()
   "Store a link to an elfeed search or entry buffer.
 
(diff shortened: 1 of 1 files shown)


tummychow:
by the way, this PR should also have the opposite effect (ie if you load elfeed without loading org-mode, org-mode should not get loaded); let me know if you observe otherwise.

skeeto:
Thanks, I think your approach makes more sense. I made one little tweak 
to use eval-after-load instead of with-eval-after-load since I still 
want to support back to Emacs 24.3.

PR #287, opened 2018-07-13
Update docstring for filter syntax to include = from README.md.

mnp:


skeeto:
Thanks! I neglected the docstring when I updated the README.

PR #312, opened 2019-04-05
In `elfeed-show-mode`, make TAB skip the link in the headers.

ieure:
Rationale: `b` opens the link and `y` saves it to the kill-ring, which
are the two primary things you’d want to do.  But since it’s a link,
you have to tab past it to get to the (different) link(s) in the
message body, which is annoying.

This PR causes TAB to skip past the message headers automatically,
preventing this annoyance.


skeeto:
This makes sense. Thanks!

skeeto:
Oh, I screwed up the issue number in 668c5af so it didn't notify, but I 
changed the function name to elfeed-show-next-link.

PR #313, opened 2019-04-06
Refactor enclosure handling.

ieure:
This completely rewrites `ELFEED-SHOW-PLAY-ENCLOSURE` and
`ELFEED-SHOW-ADD-ENCLOSURE` so they’re simpler, easier to use, and
don’t duplicate code.

 - If there’s only one enclosure, don’t prompt, just handle it.
 - Use `(interactive)` how it’s intended.  User is (maybe) prompted
   for arguments when it’s called interactively, when called from
   Lisp, the argument must be supplied & the user will never be
   prompted.
 - Remove `entry` argument, only operate on the current entry.
   This could be changed if desired, but it appears to be unused and
   is annoying to make work.
 - `ELFEED-SHOW-PLAY-ENCLOSURE` calls `ELFEED-SHOW-ADD-ENCLOSURE`
   instead of duplicating its code.
 - The FIXME in `ELFEED-SHOW-PLAY-ENCLOSURE` has been fixed.

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -396,45 +396,31 @@ offer to save a range of enclosures."
       (elfeed-show-save-enclosure-multi)
     (elfeed-show-save-enclosure-single)))
 
-(defun elfeed-show-play-enclosure (&optional entry enclosure-index)
-  "Play enclosure number ENCLOSURE-INDEX from ENTRY using emms.
-If ENTRY is nil use the elfeed-show-entry variable.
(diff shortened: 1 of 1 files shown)


skeeto:
Thanks, this is a lot cleaner. I look the liberty of fixing (and 
squashing the fix commit):

%s/maybe-get-index/maybe-prompt-index/

Merged as 2f19e61.

PR #322, opened 2019-05-30
Let elfeed--download-enclosure use curl

### LINKED ISSUE #311: Question how to let elfeed--download-enclosure use curl?
I am using elfeed to download rss feeds of podcasts. The proxy is set via:

    (setenv "ALL_PROXY" "socks5h://host:port")

I can view the feeds, but the mps files can't be downloaded. I guess it because
elfeed--download-enclosure doesn't use curl. If elfeed uses curl anyway, why not make
elfeed--download-enclosure use it too.

oystersauce8:
Fixes skeeto/elfeed#311

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -276,15 +276,8 @@ directory and saves all attachments in the chosen directory."
   "Function called to generate the filename for an enclosure.")
 
 (defun elfeed--download-enclosure (url path)
-  "Download asynchronously the enclosure from URL to PATH."
-  (if (require 'async nil :noerror)
-      (with-no-warnings
-        (async-start
-         (lambda ()
-           (url-copy-file url path t))
-         (lambda (_)
-           (message (format "%s downloaded" url)))))
-    (url-copy-file url path t)))
+  "Download the enclosure from URL to PATH."
+  (elfeed-with-fetch url (write-region (point-min) (point-max) path))
 
 (defun elfeed--get-enclosure-num (prompt entry &optional multi)
   "Ask the user with PROMPT for an enclosure number for ENTRY.


skeeto:
In principle I think this would be a good change. If the enclosure is 
downloaded asynchronously then there's really not any reason to use the 
async package, and Elfeed already knows very well how to download things 
asynchronously. The only advantage I can see is that async does the 
write-region part asynchronously, and using elfeed-with-fetch would not.

However, elfeed-with-fetch isn't a suitable interface for this. It's not 
documented as such, but it assumes the URL belongs to a feed so that it 
can track some extra state in the database (Last-Modified header, etc.). 
I admit, that whole macro is a mess and I wouldn't write it this way 
today.

Your PR is also missing a closing parenthesis. (If that was the only 
problem, I'd just fix this myself.) The Makefile makes it easy to detect 
exactly these issues. Just run "make" in the repository to build the 
whole thing, and use "make check" to run the test suite.

Using the Makefile to byte compile also reveals another issue: The 
anaphoric variable "status" isn't used. Failures are not only completely 
silent, something arbitrary will be quietly written to the destination.

So what this change should do instead is call elfeed-curl-retrieve or 
url-retrieve directly depending on the value of elfeed-use-curl, and 
properly handle errors. In the longer term perhaps elfeed-curl should 
include an option to download directly to a file instead of a buffer so 
that write-region isn't used at all.

minad:
PR has issues. Closing.

minad:
Implemented in https://github.com/emacs-elfeed/elfeed/commit/57f94d408258a4f07122a39a327283b3dd597a9a

Issue #305, opened 2019-02-19
Date ranges?


rksm:
Thanks for providing my daily news interface!

Question 1: Is there any support for date ranges? I would have expected that something like `-@1-day-ago @2-days-ago` would give me all of the posts matching inbetween those times?

Question 2: Is there any support to specify absolute dates / times? `@2019-02-19`?

skeeto:
Currently there is no support for date ranges in the search filter, 
including specific or absolute datetimes. I don't object to Elfeed 
getting such a range feature so long as it has well-designed syntax / 
notation.

rksm:
OK, I don't know how much time I'll have for this, it could be a few weeks but I would be generally be interested in drafting a PR.

So for specifying an "absolute ago" how about something that is ISO 8601 inspired (`(format-time-string "%F")` and `(format-time-string "%FT%H:%M:%S")`) together with a descriptive keyword, something like `@since-2019-02-19`, `@since-2019-02-19T13:30:00`. 

For ranges, how about reusing "--" from org ranges for something like `@2-days-ago--4-days-ago`. Semantically, @start--end. Absolute and relative date/times could be mixed like `@2-days-ago--2019-02-12` but absolute dates would loose their `since-` key.

skeeto:
I like the idea of reusing a concept from Org. Plus that should be easy 
to parse: split on "--" and pass each side to elfeed-float-time. The 
filter parser would need to be extended to have a beginning and ending 
time range, along with the parser interpreter and compiler.

rksm:
OK, will take a stab at it.

PR #324, opened 2019-06-25
Absolute dates + filter with time ranges

rksm:
This allows to specify date/time ranges as well as absolute dates/times in filters as discussed in #305.

Examples for ranges are `"@2019-06-24--2019-06-24"` or `"@5-days-ago--1-day-ago"` with the meaning that entries need to be newer than the first value and older than the second.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,11 +139,15 @@ Any component of the search string beginning with a `+` or
 a `-` is treated like a tag. `+` means the tag is required, `-` means
 the tag must not be present.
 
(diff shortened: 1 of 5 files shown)


skeeto:
Thanks! Great job with this, and especially on the tests. Those made me 
a lot more confident about the changes. The existing tests are focused 
on the core with basically no tests for the interface, but this part of 
the UI is very testable and really ought to be.

PR #330, opened 2019-07-21
fix elfeed-cleanup to indeed collapse multiple spaces

herrsimon:
multiple spaces in the middle of a string were not collapsed before.


skeeto:
Your PR was a little bit wrong, so I tweaked it and merged as 7bdccb1. 
Your change didn't chain the regexp replacements, so the first 
replacement was lost leading to incorrect output. It also didn't fix one 
of the tests that expected the internal whitespace to be preserved (why 
did I do that?). You can run the test suite yourself with "make check".

The only change that was actually needed was for the first regexp to 
include a space so that it gets collapsed just like the other 
whitespace. I also took the opportunity to add the other missing 
whitespace characters.

Thanks!

herrsimon:
Ok, no problem, thanks again for this great package!

PR #332, opened 2019-08-09
Refine behavior of "c", "<", and ">" keybindings.

cowile:
Make filter clearing update the search view.

Make ">" go to the last RSS article rather than the end of the buffer by backing up two lines.


skeeto:
Good ideas, thanks!

PR #344, opened 2019-11-14
Add user option elfeed-search-remain-on-entry

rnkn:
When non-nil, keep point at entry after performing a command.

When nil, move to next entry.


rnkn:
Wait, sorry I mixed up the use-region-p... please stand by...

rnkn:
And fixed. Sorry for the noise.

skeeto:
Great PR, thanks!

PR #346, opened 2019-11-16
Ensure bookmarks are loaded for elfeed-search

rnkn:
According to the docstring of bookmark-jump:

  You may have a problem using this function if the value of
  variable ‘bookmark-alist’ is nil. If that happens, you need to
  load in some bookmarks. See help on function ‘bookmark-load’ for
  more about this.

So e.g. binding a key to jump to an elfeed bookmark will cause an
error without first loading bookmarks.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -8,6 +8,7 @@
 (require 'browse-url)
 (require 'wid-edit) ; widget-inactive face
 (require 'bookmark)
+(bookmark-maybe-load-default-file)
 
 (require 'elfeed)
 (require 'elfeed-db)


skeeto:
Thanks, I didn't know about this. Merged.

IMHO, it should really be bookmark's responsibility to call this 
function itself when it's needed rather than produce an error. In 
contrast, Elfeed has an equivalent function, elfeed-db-ensure. It's 
called automatically at the last moment, just before the database is 
needed.

rnkn:
Agreed. It seems a bit odd that bookmark-jump has the disclaimer rather than just call bookmark-maybe-load-default-file itself...

Thanks for the merge!

Issue #347, opened 2019-12-06
Excluding feeds ?


aagon:
Hello Mr. Wellons,

First of all, thank you for this wonderful package. I (almost) do not need to go open youtube.com in a web browser anymore.

Certain feeds tend to have a lot more output than others, thus cluttering the elfeed-search buffer. There is a way to display only entries coming from certain feeds ('='), but none to display entries **not** coming from certain feeds.

One solution would be to have separate tags for each feed, and then use -. But I found the implementation of the "exclude-feed" filter quite easy.

I will open a PR immediately, and comment more extensively on my changes

Thank you again,

Aymeric Agon-Rambosson

skeeto:
Fixed by #348.

PR #348, opened 2019-12-06
Implemented "exclude feed" functionality.

aagon:
This allows to specify several feeds to be excluded from the elfeed-search buffer, as discussed in #347.

I have modified 4 functions :
- elfeed-search-parse-filter
- elfeed-search-unparse-filter
- elfeed-search-filter
- elfeed-search-compile-filter

I have simply added a special character ('~') in the first two functions, and added a condition in the last two : similarly to =, if the entry's feed name or id matches **any** the ~-prepended strings, it is excluded.

The modified repo builds correctly (make gives no errors) and passes all the tests (make check is satisfied), although I haven't included any new ones. I have tested the package manually on the new features, though, and it works : the amount of entries with ~%s plus the amount of entries with =%s is always equal to the amount without any filters (tested on a dozen strings).

I hope you find this satisfactory.

Aymeric Agon-Rambosson

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,10 @@ A component beginning with a `=` is a regular expression matching the
 entry's feed (title or URL). Only entries belonging to a feed that
 match at least one of the `=` expressions will be shown.
 
+A component beginning with a `~` is a regular expression matching the
+entry's feed (title or URL). Only entries belonging to a feed that
+match none of the `~` expressions will be shown.
+
 All other components are treated as a regular expression, and only
 entries matching it (title or URL) will be shown.
 
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -380,7 +380,8 @@ The customization `elfeed-search-date-format' sets the formatting."
         (matches ())
         (not-matches ())
         (limit nil)
-        (feeds ()))
+        (feeds ())
+        (not-feeds ()))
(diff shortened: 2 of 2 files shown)


skeeto:
This looks great, thanks! I'm ready to merge it except for one small 
thing. Could you set indent-tabs-mode to nil, M-x whitespace-cleanup, 
then amend / squash your PR with that change? I could do this myself, 
but I want a tidy history and it would be nice for GitHub to properly 
recognize the merge / acceptance.

I'd also like to have this feature mentioned in README.md just after the 
paragraph about =. If you don't want to do this, that's fine. I can just 
add it in another commit after I merge the PR.

Thanks for the solid PR!

aagon:
I did the modifications you asked, the tabs have been removed. I have also added the documentation to the README.

About that, I am not a native English speaker, but should we not say : "Only entries belonging to a feed that match**es** at least / none etc...". I have sticked to the model, but I feel it should be modified.

I can amend the commit with that change as well.

skeeto:
Perfect, thanks!

You're absolutely right about the README. At first glance it might seem 
that the verb "match" goes with "entries" in which case it would be 
correct (both plural), but it actually goes with "feed", singular. It 
slipped by me for that reason. Since it's simple, I just fixed it in 
352f58a.

PR #349, opened 2019-12-18
Add a command keybinding to kill link URL into kill-ring

stardiviner:
In many situation, I need to copy out the link URL to other places.


skeeto:
Looks good, thanks!

PR #350, opened 2019-12-25
support copy image link with shr command

stardiviner:
Improve commit "a99122e" copy command.

----

#

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -459,8 +459,9 @@ Prompts for ENCLOSURE-INDEX when called interactively."
   (interactive)
   (let ((url (or (elfeed-get-link-at-point)
                  (elfeed-get-url-at-point))))
-    (kill-new url)
-    (message url)))
+    (if url
+        (progn (kill-new url) (message url))
+      (call-interactively 'shr-maybe-probe-and-copy-url))))
 
 (provide 'elfeed-show)
 


skeeto:
Thanks!

alphapapa:
I'm not sure this is a good idea.  `shr-probe-and-copy-url` is:

```el
(defun shr-probe-and-copy-url (url)
  "Copy the URL under point to the kill ring.
Like `shr-copy-url', but additionally fetch URL and use its
redirection destination if it has one."
  (interactive (list (shr-url-at-point current-prefix-arg)))
  (if url (shr-probe-url url #'shr-copy-url)
    (shr-copy-url url)))
```
So it FETCHES the URL.  Simply copying a URL to the kill ring should not FETCH the URL.

alphapapa:
In case it's not clear:
```el
(defun shr-probe-url (url cont)
  "Pass URL's redirect destination to CONT, if it has one.
CONT should be a function of one argument, the redirect
destination URL.  If URL is not redirected, then CONT is never
called."
  (interactive "P")
  (url-retrieve
   url (lambda (a)
         (pcase a
           (`(:redirect ,destination . ,_)
            ;; Remove common tracking junk from the URL.
            (funcall cont (replace-regexp-in-string
                           ".utm_.*" "" destination)))))
   nil t))
```

stardiviner:
@alphapapa You're right, I will update it to `shr-copy-url`. Thanks.

skeeto:
Thanks, @alphapapa.

PR #354, opened 2020-01-19
Add an option to start an timer for auto updating elfeed when Emacs idle.

stardiviner:


skeeto:
Thanks. I'll accept this with changes, preferably squashed into this 
commit:

* I don't want this enabled by default, and definitely not to such an 
aggressive rate of every 10 minutes. (IMHO, feed readers should not be 
making requests to a server more than once an hour, and most feeds 
publish slow enough to be fetched much less often than that.)

* Since you have the timer set to repeat, and the timer function 
(elfeed-update) spawns a new timer, won't this continuously create 
redundant timers? That would be a disaster if left running for long. 
More on this below.

* The :type and :safe properties must reflect that it could be either 
nil or a number. Also, anything smaller than 1 is not a safe setting.

* There should be a line break at the end of the first sentence in the 
docstring. Conventionally the first line of a docstring "should be a 
complete sentence that gives a summary description of the object."

* The final argument to run-with-idle-timer should use a sharp quote.

* In elfeed-auto-idle-update, since there's no "else" to the if 
statement I'd prefer "when".

Related to the repeating timer issue: I don't think the user interface 
has been fleshed out. I *do* like that auto updates don't start until 
the user manually runs elfeed-update. Every time there's an update, even 
automatically, the timer duration should be updated to the current 
configuration. I'm not actually sure how to do this properly with an 
idle timer. If nil, the timer should be canceled. (This all means 
keeping track of the timer object.)

Further, when the timer function runs, it should first check that the 
configuration variable hasn't been set to nil before updating. Setting 
it to nil will then automatically "cancel" any pending auto update.

stardiviner:
> Related to the repeating timer issue: I don't think the user interface
has been fleshed out. I *do* like that auto updates don't start until
the user manually runs elfeed-update. Every time there's an update, even
automatically, the timer duration should be updated to the current
configuration. I'm not actually sure how to do this properly with an
idle timer. If nil, the timer should be canceled. (This all means
keeping track of the timer object.)

I agree with your opinion, I have not considered a lot about this. Looks my programming skill need more deep thought. :) Thanks @skeeto . I will close this PR.

PR #356, opened 2020-01-23
WIP: Add makem.sh, test.yml, minor adaptations to test files

alphapapa:
Hi Chris,

Here's an example of using `makem.sh` for testing and linting Elfeed.  The easiest way to do it is:

1.  Checkout this PR.
2.  Run:
```sh
mkdir sandbox
./makem.sh -vv --sandbox-dir sandbox --install-deps --install-linters
./makem.sh -vv --sandbox-dir sandbox all  # or "test" or "lint" instead of "all"
```
You can also use `--sandbox` instead of `--sandbox-dir sandbox` to install current versions of dependencies in a temporary sandbox in `/tmp`, which is deleted on exit.

I moved a few function and variable defs in the test files so they load properly with `makem.sh` (since it doesn't guarantee the order in which the test files are loaded), and added headers to declare the external dependencies the tests require.

Having done this, all the tests pass except for 2.  There are no byte-compile errors, so I'm not sure if these 2 tests are broken by my changes.

I also added a GitHub Actions CI file testing all Emacs versions from 24.3 to current snapshot.  You can see the results here: https://github.com/alphapapa/elfeed/runs/404412585

Note that I did not overwrite the Makefile with the one from the `makem.sh` repo, which just runs the script anyway.

I don't necessarily expect you to merge this PR.  I'm just making it to show you how `makem.sh` could work for this package.


alphapapa:
BTW, I just added https://github.com/alphapapa/makem.sh/commit/2c6937aef7b12f091352483d9bc03a5dc64d0a7a so that the `--sandbox-dir` persistent sandbox option uses Emacs version-specific subdirectories, so you can easily use the `--emacs` option with it to test with multiple Emacs versions while persisting each version's set of installed packages.

skeeto:
Thanks for the illustration, alphapapa. I'm currently quite happy with 
my Makefile-driven build and tests, so I'm going to stick with that.

I was alarmed by the tests not passing, so I dug into it. Turns out 
the elfeed-search tests were never being run, and they never worked in 
the first place. When I merged that PR I hadn't noticed the new tests 
weren't integrated, and so when "make check" passed I assumed the tests 
worked. That's now fixed, so thanks for revealing that testing bug!

PR #359, opened 2020-02-02
Add support for bookmarks in entries

### LINKED ISSUE #358: Bookmark support for *elfeed-entry*/elfeed-show?
I was reading a very big article and thought i'd bookmark where i'd stopped and read it later with `bookmark-set`, but it says `bookmark-buffer-file-name: Buffer not visiting a file or directory` when i try to create it. I've heard bookmarks work with searches, so maybe they should also work with entries, if possible?

vaartis:
Closes #358

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -9,6 +9,8 @@
 (require 'url-parse)
 (require 'browse-url)
 (require 'message) ; faces
+(require 'bookmark)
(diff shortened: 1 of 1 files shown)


skeeto:
Thanks!

alphapapa:
Cool, thanks @vaartis!  It's always nice when a great feature is only a few lines of code!

PR #364, opened 2020-02-09
Add defcustoms for elfeed-show-entry- functions

### LINKED ISSUE #361: elfeed-show-entry-switch should be a defcustom
I think `elfeed-show-entry-switch` should be a `defcustom` so that it would be more discoverable and to indicate that it is intended to be customized.  :)

alphapapa:
Fixes #361.


skeeto:
Perfect, thanks!

PR #366, opened 2020-02-15
Excerpts - TAB unfolding of entries in the search buffer

alphapapa:
Hi Chris,

This is a WIP implementation of "excerpts".  Two keys are bound in the search buffer: `e`, which toggles an excerpt for the entry at point, and `E`, which toggles as many as fit in the window (and displays the first unread entry even if it doesn't fit).

Rather than insert text into the buffer, which interferes with the calculation of entry positions in the buffer, overlays are used.  This presents its own minor issues; please see the comments in the code for details.

Despite the minor issues, I find it very useful to quickly view entries in a single window, several at a time (of course, it works best for shorter entries).  Here's how I'm testing it on my end, using `makem.sh` (I know you have sandbox code in the Makefile, but this is easier for me.  :):

```
$ ./makem.sh --sandbox .sandbox/excerpts --install-deps --install-linters
$ cat >>.sandbox/excerpts/26.3/init.el <<EOF
(setf elfeed-db-directory (expand-file-name "elfeed" user-emacs-directory)
      elfeed-feeds '(("https://planet.emacslife.com/atom.xml" Emacs)))

(find-file user-init-file)
(split-window-right)
(elfeed)
EOF
$ ./makem.sh --sandbox .sandbox/excerpts interactive

# Then press "G" to refresh feeds, and then repeatedly press "E" to
# view entries in the search buffer.
```

It's not quite perfect (see comments in code), but it works pretty well and is useful.  Maybe one of us will come up with some ideas about how to polish the issues.

I've left some WIP commits in the branch, mainly to show the first commit that used text insertion rather than overlays, in case we want to go that route instead (see comments).  If the branch is acceptable, or becomes so, it can be squashed before merging.

Please let me know what you think.  Thanks.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -962,6 +962,66 @@ Sets the :title key of the feed's metadata. See `elfeed-meta'."
 (add-to-list 'desktop-buffer-mode-handlers
              '(elfeed-search-mode . elfeed-search-desktop-restore))
 
+;;;; Inline view
+
+;; This section implements inline viewing of entries in the search
+;; buffer.
+
+(defmacro elfeed-search-at-entry (entry &rest body)
+  "If ENTRY is in search buffer, eval BODY with point at it."
+  (declare (indent defun))
+  `(let ((n (cl-position ,entry elfeed-search-entries)))
+     (when n
+       (elfeed-goto-line (+ elfeed-search--offset n))
+       ,@body)))
+
+(defun elfeed-search-inline-toggle ()
+  "Toggle display of current or selected entries inline."
+  (interactive)
+  (cl-labels
+      ((inline-bounds
+        ;; Return (beg . end) of ENTRY's inline entry text.
+        (entry) (elfeed-search-at-entry entry
+                  (let ((beg (next-property-pos (point) :elfeed-entry-inline)))
+                    (when beg
+                      (cons beg (or (next-property-pos beg :elfeed-entry-inline)
+                                    (save-excursion
+                                      (goto-char beg)
+                                      (forward-line 1)
+                                      (point))))))))
+       (next-property-pos
+        ;; Return next position from POS where PROPERTY changes (and matches VALUE, if set).
+        (pos property &optional value)
+        (let ((pos (next-single-property-change pos property)))
+          (when pos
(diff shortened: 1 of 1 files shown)


alphapapa:
@skeeto FYI.  :)

alphapapa:
@skeeto Ping?  :)

minad:
@alphapapa Not sure if you still use Elfeed - I tried this patch and it is really nice. However images are not loaded correctly since you use an overlay. I am not sure how we could solve this with the shr library.

alphapapa:
Hi @minad, I don't regularly read RSS feeds lately, but I'd still like to see this get merged someday. I'm guessing it will need someone to fork and maintain Elfeed. Anyway, I guess using it with images would require loading the image to learn its dimensions and including that in the calculation using pixel text height, or something like that.

minad:
@alphapapa 

Thanks for hearing back from you! I hope you're doing well.

> I'm guessing it will need someone to fork and maintain Elfeed. 

I am maintaining it now, together with Ihor and Karthik.

Regarding image loading the problem is that shr loads the image lazily and inserts it then as overlay or display property. But this doesn't work if the rendered text is already put in an overlay. Shr assumes that the rendered text is inserted directly in a buffer.

alphapapa:
> @alphapapa 
> 
> Thanks for hearing back from you! I hope you're doing well.

Thanks. I'm doing okay, I just am busy with work, and I don't have as much energy for Emacs things in my free time as I used to. But every weekend I feel like I ought to tackle a few issues on my projects... ;)

> 
> > I'm guessing it will need someone to fork and maintain Elfeed. 
> 
> I am maintaining it now, together with Ihor and Karthik.

Great! 

> 
> Regarding image loading the problem is that shr loads the image lazily and inserts it then as overlay or display property. But this doesn't work if the rendered text is already put in an overlay. Shr assumes that the rendered text is inserted directly in a buffer.

Hm, that sounds like it would require some trickery...

minad:
> Hm, that sounds like it would require some trickery...

I thought maybe one needs a separate buffer, and then copy the content back to the overlay. Would be quite a hack unfortunately.

PR #386, opened 2020-06-28
Advertise support for Atom in HTTP Accept header

tomfitzhenry:
Per https://tools.ietf.org/html/rfc7231#section-5.3.2 HTTP clients may
advertise support for content types via the Accept header.

Some HTTP servers that serve Atom require this, e.g. https://developer.github.com/v3/activity/feeds

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -177,7 +177,8 @@ of `elfeed-use-curl'."
          (let* ((feed (elfeed-db-get-feed url))
                 (last-modified (elfeed-meta feed :last-modified))
                 (etag (elfeed-meta feed :etag))
-                (headers `(("User-Agent" . ,elfeed-user-agent))))
+                (headers `(("User-Agent" . ,elfeed-user-agent)
+                           ("Accept" . "application/atom+xml"))))
            (when etag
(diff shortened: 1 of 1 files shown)


skeeto:
This Accept header is too narrow since Elfeed also accepts RSS. A server 
that only serves RSS would be right to reject a request from Elfeed with 
this header, and that's not what we want. So, at minimum, Elfeed should 
specify both.

However, Elfeed already does this: "Accept: */*". That's too broad, so 
technically GitHub could respond with some stupid default like JSON. 
Fortunately the GitHub documentation is wrong and that endpoint serves 
an Atom feed in response to */*, or even with no Accept header at all.

I looked at a few other feed readers, and only found two with precise 
Accept headers. Liferea:

Accept: application/atom+xml,application/xml;q=0.9,text/xml;q=0.8,*/*;q=0.7

QuiteRSS:

Accept: application/atom+xml,application/rss+xml;q=0.9,application/xml;q=0.8,text/xml;q=0.7,*/*;q=0.6

I'm not thrilled at the idea of using such a complex header, but I 
suppose it's technically a little more correct.

PR #389, opened 2020-07-16
Refactor: leverage cl-destructuring-bind

progfolio:
Simplify series of `plist-get` assignments with `cl-destructuring-bind`.


skeeto:
Thanks! I didn't know about this feature, and, even more, I'm surprised 
it's been there since at least Emacs 24.3.

progfolio:
You're welcome! I was surprised to learn about it myself. It's a useful pattern.

alphapapa:
See also `pcase-let*`, which is also built-in to Emacs and can be even more concise and flexible.

PR #397, opened 2020-11-27
Add support for local feeds.

### LINKED ISSUE #300: Allow fetching from local files (file:// "protocol")
Hi,

feature request: allow fetching from `file://` URLs. I have a script that scrapes websites that don't have feeds and builds Atom feeds from them that are output to a temporary directory. elfeed fails to fetch such feeds with an error message:

~~~~
Error running timer ‘elfeed-curl--call-callback’: (wrong-type-argument number-or-marker-p nil)
~~~~

`curl` itself does support fetching via `file://` (I tested it on the commandline). Currently, I work around the problem by using a web server on `localhost`, but that's just awkward.

justinabrahms:
Fixes #300

### DIFF
--- a/elfeed-curl.el
+++ b/elfeed-curl.el
@@ -338,12 +338,17 @@ DATA is the content to include in the request."
       (unwind-protect
           (progn
             (elfeed-curl--prepare-response url n)
-            (if (and (>= elfeed-curl-status-code 400)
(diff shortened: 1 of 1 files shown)


skeeto:
That was simple, thanks! Merged as cc9d3b2.

PR #398, opened 2020-12-01
Fix up elfeed-show-next/prev

br0cco1i:
These functions expected `#'elfeed-search-show-entry` would move forward one
line. However, when `elfeed-search-remain-on-entry` is set to `t`, it won't, so we
need to do it manually.


skeeto:
Good catch, thanks!

PR #402, opened 2020-12-20
No longer query users about running curl processes

sinic:
This is very minor, but I feel users shouldn't be queried about ephemeral processes like the asynchronous instances of `curl` spawned by Elfeed.

If the processes are old, they will probably time out and there is no harm in killing them. If they are young, then the user's plans presumably changed right after calling `elfeed-update`, and a query would only serve as an obstruction.

(You may also wish to call `set-process-sentinel` in the line after the one I added, for consistency.)


skeeto:
Thanks! I was unaware of this issue since 1) the only time I exit Emacs 
is when I shut down my computer, and 2) I don't think I've ever exited 
during feed updates.

PR #404, opened 2020-12-28
Use property align-to indent title pixel width

manateelazycat:
If title include CJK font, title won't alignment like export, because Chinese char is not same width as English char.

This patch use align-to force title indent width special pixel.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -346,20 +346,19 @@ The customization `elfeed-search-date-format' sets the formatting."
                     (lambda (s) (propertize s 'face 'elfeed-search-tag-face))
                     tags ","))
          (title-width (- (window-width) 10 elfeed-search-trailing-width))
-         (title-column (elfeed-format-column
-                        title (elfeed-clamp
-                               elfeed-search-title-min-width
-                               title-width
-                               elfeed-search-title-max-width)
-                        :left)))
+         (title-format-width (elfeed-clamp
+                              elfeed-search-title-min-width
+                              title-width
+                              elfeed-search-title-max-width))
+         (title-column (elfeed-format-column title title-format-width :left))
+         (align-width (if (> (window-width) (/ (* 2 (frame-width)) 3))
+                          title-width
+                        (+ (string-width date) 5 title-format-width))))
     (insert (propertize date 'face 'elfeed-search-date-face) " ")
     (insert (propertize title-column 'face title-faces 'kbd-help title) " ")
 
     ;; Use property `align-to' indent title pixel width, otherwise CJK title not alignment like expect.
-    (put-text-property
-     (point) (- (point) 1)
-     'display
-     (elfeed-indent-pixel (* title-width (window-font-width))))
+    (put-text-property (point) (- (point) 1) 'display (elfeed-indent-pixel (* align-width (window-font-width))))
 
     (when feed-title
       (insert (propertize feed-title 'face 'elfeed-search-feed-face) " "))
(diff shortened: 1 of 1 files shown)


manateelazycat:
This screenshot is status before apply patch, CJK is not alignment.

manateelazycat:
Here is screenshot before apply patch https://ibb.co/RDzxScK

CJK char indent after apply patch https://ibb.co/pJz4z74

skeeto:
I'm indeed seeing improved behavior when the frame is wide:

before
after

But no improvement when the frame is thin, where entry titles are truncated:

before-thin
after-thin

I also don't like that the space between entry title and feed title is gone, but that could probably be fixed by prefixing titles with a space.

I tested with <https://www.tofugu.com/feed> (the first thing I could find), but feel free to suggest any other feeds for testing.

chen-chao:
> But no improvement when the frame is thin, where entry titles are truncated

This is because `elfeed-indent-pixel` position is unfortunately back in the title characters when `window-width` is small (it ignores `elfeed-search-title-min-width` and `elfeed-search-title-max-width` actually).

A value that is complicated but near to the original behavior is:

```
(+ date-width 2 (max elfeed-search-title-min-width (min title-width elfeed-search-title-max-width)))
``` 

The pixel length of multi-bytes CJK characters is less than the pixel length of equal-bytes spaces (i.e. CJK titles are shorter). So this should be fine.

manateelazycat:
@skeeto I commit new patch to make align-to property works with thin window.

chuxubank:
@manateelazycat The newest commit still have bug on wide screen
<details>
    <summary>Details</summary>
image
</details>

manateelazycat:
> @manateelazycat The newest commit still have defect on wide screen
> 
> Details

Done

manateelazycat:
@skeeto Can you merge this patch? Thanks.

skeeto:
In the current version on my machine non-CJK entries are now staggered, 
and CJK entries still aren't working right at some widths. The position of 
the feed title varies greatly depending on the font weight (i.e. unread 
vs. read), and at thinner widths entry titles are cut off well before the 
feed title. I'm testing using "make virtual" and just accepting the 
default fonts, etc.

In order for me to merge the current behavior for non-CJK must be 
absolutely unaffected, "make check" must be completely clean (no 
warnings), and the new CJK behavior must be objectively better (no 
regressions).

I appreciate your effort, Andy, but I suspect this is an essentially 
unresolvable issue related to the way Emacs renders certain fonts. Elfeed 
already uses string-width (rather than string length) to determine the 
display width for titles and such, and so everything is already aligned 
correctly for fonts where wide CJK glyphs are exactly two columns wide. 
When I run Elfeed in a terminal, CJK is already lined up correctly since 
my terminal is rendering the glyphs in the straightforward way. The pixel 
alignment stuff will never work in terminal Emacs, and so it will always 
be complete.

Here's a similar struggle with lining up tables in Org:

https://emacs.stackexchange.com/questions/19808/align-columns-in-org-mode-table-with-japanese-fonts

yantar92:
FYI. I am using the following function to determine displayed pixel
width of a string in graphical Emacs. Hope it can be helpful.

(defun string-display-pixel-width (string &optional mode)
  "Calculate pixel width of STRING.
Optional MODE specifies major mode used for display."
  (with-temp-buffer
    (with-silent-modifications
      (setf (buffer-string) string))
    (when (fboundp mode)
      (funcall mode)
      (font-lock-fontify-buffer))
    (if (get-buffer-window (current-buffer))
	(car (window-text-pixel-size nil (line-beginning-position) (point)))
      (set-window-buffer nil (current-buffer))
      (car (window-text-pixel-size nil (line-beginning-position) (point))))))

PR #414, opened 2021-02-12
Fix comment and last-entry behaviour

utkarsh181:
I think you should also swap the order of these two functions.


skeeto:
Good catch, thanks!

PR #423, opened 2021-05-28
{tag,untag,toggle}-all use multiple tags

Accacio:


minad:
Implemented in https://github.com/emacs-elfeed/elfeed/commit/8b6f058c360c64d130ac8447faec664141ac3e9f

Issue #32, opened 2014-06-28
Better completition in search dialog


Fuco1:
I think it would be nice to have better completition in the search dialog (s or S).  Specifically, tags could be autocompleted (if the word starts with `+`/`-`), also dates (with @) format, for example `@2-d` could expand to `days` etc.

To achieve this, it would be necessary to get a list of all existing tags. Is there a simple way to do this? I could try to hack the completition "engine" myself, but I don't understand much how the elfeed database works.

Thanks

Fuco1:
I'm going to close this to keep the list of my issues under control.  I do not plan to work on this anymore.  Thanks for understanding!

PR #424, opened 2021-05-29
Add autocompletions for tags & titles in elfeed-search buffer.

SqrtMinusOne:
This adds notmuch-like autocompletions in the `*elfeed-search*` buffer. I've seen some discussion in #32 a few years ago, but didn't find anything like that implemented.

It works for:
- `+...` - tags
- `-...` - tags
- `=...` - titles

That's a subset of the possible filter syntax but in my opinion much better than nothing.

The variable `elfeed-search-completion-enabled` enables autocompletion. `elfeed-search-completion-cache` enables caching. I've started using elfeed rather recently, so I am not sure how fast would it be to fetch tags and titles from DB at every request. Therefore it's optional and enabled by default.

Built-in `completing-read` and `completing-read-multiple` from `crm.el` are used, so the completion should work with any popular completion framework.

### DIFF
--- a/elfeed-db.el
+++ b/elfeed-db.el
@@ -266,6 +266,16 @@ The FEED-OR-ID may be a feed struct or a feed ID (url)."
       (maphash (lambda (k _) (push k tags)) table)
       (cl-sort tags #'string< :key #'symbol-name))))
 
+(defun elfeed-db-get-all-titles ()
+  "Return a list of all titles currently in the database."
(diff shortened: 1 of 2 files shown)


minad:
Implemented basic completion in https://github.com/emacs-elfeed/elfeed/commit/8b6f058c360c64d130ac8447faec664141ac3e9f and https://github.com/emacs-elfeed/elfeed/commit/28dad74e2cecf351f072b06d422479e4693db7b3.

PR #426, opened 2021-06-06
Ensure elfeed-search-browse-url stays in elfeed buffer (fixes #425)

### LINKED ISSUE #425: How to mark entry as read after browse url with eww?
I didn't find if there a hook about browse url .

tsdh:
If some emacs internal browser such as `eww` is invoked by `browse-url`, the
remainder of the function ran in the `eww` buffer which didn't make sense and
didn't visually mark the entry as read.


skeeto:
Thanks! I made a tweak to use with-current-buffer instead of set-buffer, 
then merged as 243add9.

PR #430, opened 2021-07-06
Assign `list-buffers-directory` to the search filter

nagy:
This displays the search filter for elfeed-search buffers in
`M-x list-buffers` instead of showing an empty path column.


alphapapa:
Looks like someone discovered an Emacs feature: https://github.com/Kungsgeten/org-brain/pull/370  ;)

PR #438, opened 2021-08-22
Support display tags with customized icons

stardiviner:
Improve elfeed looking by display tags with icons.

Here is the screenshot of the effect:

image


stardiviner:
@skeeto gentle ping....

Thaodan:
Maybe move this to an external package or optional feature? I've thought of doing something like this either your way or by using faces if possible.
However some might not like it.

stardiviner:
If you have ideas to implement similar things. It's great. Using external package for this simple feature is not a good idea. I'm currently using a local fork of elfeed. So I will wait for your implementation. Now I close this PR.

Thaodan:
As suggested you could just make this optional or implementing
format-strings, I was thinking of something like
gnus-summary-line-format (C-h v gnus-summary-line-format RET).

I just suggest this I don't necessarily have the time to implement such a feature.

stardiviner:
Oh, I misunderstand your meaning. I will consider to make it optional and `gnus-summary-line-format`. When I updated, will update this PR. Thanks.

brongulus:
Hey, any updates regarding this PR, would love to see this as part of the official repo.

PR #439, opened 2021-08-22
Update example link

idanko:
The `http://planet.emacsen.org/atom.xml` no longer exists, update to a new one `https://planet.emacslife.com/atom.xml`


skeeto:
Thanks! I probably should have updated this when the feed originally 
disappeared a few years ago, but I forgot all about it being in the 
README.

PR #440, opened 2021-08-25
Faster elfeed-search-update

shitikanth:
Save the value of the corresponding elfeed-entry in text-property and
only call elfeed-search-print-entry-function when the entry has changed
or we have encountered a new entry.

On my system, with ~7000 elfeed-search-entries this reduces the time of
elfeed-search-update with a small number of changes from ~0.70s to ~0.03s.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -360,16 +360,12 @@ The customization `elfeed-search-date-format' sets the formatting."
                                title-width
                                elfeed-search-title-max-width)
                         :left)))
-    (let ((beg (point)))
-      (insert (propertize date 'face 'elfeed-search-date-face) " ")
-      (insert (propertize title-column 'face title-faces 'kbd-help title) " ")
-      (when feed-title
-	(insert (propertize feed-title 'face 'elfeed-search-feed-face) " "))
-      (when tags
-	(insert "(" tags-str ")"))
-      (add-text-properties
-       beg (point)
-       `(saved-elfeed-entry ,(copy-elfeed-entry entry))))))
+    (insert (propertize date 'face 'elfeed-search-date-face) " ")
+    (insert (propertize title-column 'face title-faces 'kbd-help title) " ")
+    (when feed-title
+      (insert (propertize feed-title 'face 'elfeed-search-feed-face) " "))
+    (when tags
+      (insert "(" tags-str ")"))))
 
 (defun elfeed-search-parse-filter (filter)
   "Parse the elements of a search filter into a plist."
@@ -703,15 +699,6 @@ expression, matching against entry link, title, and feed title."
                              line))
          (move-to-column column)))))
 
-(defun elfeed-entries-compare (entry-a entry-b)
-  (let* ((a (elfeed-entry-id entry-a))
-	 (b (elfeed-entry-id entry-b))
-	 (date-a (elfeed-entry-date entry-a))
-         (date-b (elfeed-entry-date entry-b)))
-    (if (= date-a date-b)
-        (string< (prin1-to-string b) (prin1-to-string a))
-      (> date-a date-b))))
-
 (defun elfeed-search-update (&optional force reset)
   "Update the elfeed-search buffer listing to match the database.
 When FORCE is non-nil, redraw even when the database hasn't changed."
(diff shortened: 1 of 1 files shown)


shitikanth:
In current state, if window-width or any parameter for the elfeed-search-print-entry-function has changed in the meantime, the un-updated rows and updated rows will look inconsistent. For such a situation, I have added an option for a 'reset argument which clears the buffers before proceeding.

elfeed-search.el
```diff
@@ -360,12 +360,16 @@ The customization `elfeed-search-date-format' sets the formatting."
...
-    (when tags
-      (insert "(" tags-str ")"))))
+    (let ((beg (point)))
+      (insert (propertize date 'face 'elfeed-search-date-face) " ")
+      (insert (propertize title-column 'face title-faces 'kbd-help title) " ")
+      (when feed-title
+	(insert (propertize feed-title 'face 'elfeed-search-feed-face) " "))
+      (when tags
+	(insert "(" tags-str ")"))
+      (add-text-properties
```
alphapapa, line 370: You should probably use `put-text-property` here, since you're only setting one property.

elfeed-search.el
```diff
@@ -699,22 +703,52 @@ expression, matching against entry link, title, and feed title."
                              line))
          (move-to-column column)))))
 
-(defun elfeed-search-update (&optional force)
+(defun elfeed-entries-compare (entry-a entry-b)
```
alphapapa, line 706: Please reorder these functions so the diff can be more cleanly read (see how it shows the `elfeed-search-update` function's first line being replaced, but the new line is far below it in the diff).

elfeed-search.el
```diff
@@ -699,22 +703,52 @@ expression, matching against entry link, title, and feed title."
...
-          (erase-buffer)
-          (elfeed-search--update-list)
-          (dolist (entry elfeed-search-entries)
-            (funcall elfeed-search-print-entry-function entry)
-            (insert "\n"))
-          (setf elfeed-search-last-update (float-time))))
+	(let ((inhibit-read-only t)
+	      (standard-output (current-buffer))
+	      (sorter (or elfeed-search-sort-function #'elfeed-entries-compare)))
+	  (when reset
```
alphapapa, line 726: Is it necessary to have both `force` and `reset` flags?

elfeed-search.el
```diff
@@ -699,22 +703,52 @@ expression, matching against entry link, title, and feed title."
...
+	    (erase-buffer))
+	  (elfeed-search--update-list)
+	  (goto-char (point-min))
+	  (dolist (entry elfeed-search-entries)
+	    (while
+		(let ((saved-entry (get-text-property (point) 'saved-elfeed-entry)))
+		  (cond
+		   ;; no change in entry
```
alphapapa, line 734: This comment is a bit redundant, considering the simple `equal...` below.  :)

elfeed-search.el
```diff
@@ -699,22 +703,52 @@ expression, matching against entry link, title, and feed title."
...
+		   ((equal entry saved-entry)
+		    (forward-line 1)
+		    nil)
+		   ;; saved-entry sorts after entry, insert and moved on
```
alphapapa, line 738: This comment is not very clear.  Could you rewrite it as a sentence?

elfeed-search.el
```diff
@@ -699,22 +703,52 @@ expression, matching against entry link, title, and feed title."
...
+		   ((or (not saved-entry)
+			(funcall sorter entry saved-entry))
+		    (funcall elfeed-search-print-entry-function entry)
+		    (insert "\n")
+		    nil)
+		   ;; found an entry that sorts before id,
```
alphapapa, line 744: This comment could be more succinct, like "Entry sorts before ID: delete it."

alphapapa (commented):
Of course, I'm not Chris, but maybe I can save him a little time by suggesting a few things.

In addition to the specific comments, please `untabify` your changes and be sure they are indented properly, because several of these forms aren't.

minad:
@shitikanth Hi, thank you for this PR! I merged it with a few modifications. In case you're still an Elfeed user, I would appreciate if you would give it a through check.

EDIT: Reverted again, for now. See https://github.com/emacs-elfeed/elfeed/commit/06faad4092c6cbaeca2224ab3995ec42fab02369 for the reasoning.

PR #442, opened 2021-09-05
dereference entry content when attaching to org link

dellison:
For elfeed entries like `elfeed-show-entry`, the "content" is actually a reference to the content in db rather than the content itself. Dereferencing it here when creating the link means that an org capture template can use `%:elfeed-entry-content` for the content as one would expect.


minad:
This has already been implemented:

https://github.com/emacs-elfeed/elfeed/blob/aad8d96818d5a68136fbd534897bedaf65122f32/elfeed-link.el#L89-L94

Issue #365, opened 2020-02-11
Handling HTML in feed titles


ahacop:
First, thanks very much for the great software.

A handful of the feeds that I follow use HTML tags and entities in the title of an article. In order to decode the HTML, in my config I've added a decoding function and modified a couple of elfeed functions (`elfeed-search-print-entry--default` and `elfeed-show-refresh--mail-style`) to make use of the decoding function. I've never written elisp, so there is probably a better way to do this. In any case, maybe someone out there will find this information useful.

```elisp
(defun decode-entities/libxml (html)
   (unless (fboundp 'libxml-parse-html-region)
     (error "This function requires Emacs to be compiled with libxml2"))
   (with-temp-buffer
     (insert html)
     (let* ((document (libxml-parse-html-region (point-min) (point-max)))
            (decoded-string (dom-texts document)))
       decoded-string)))
```

skeeto:
HTML in titles is one place where Elfeed falls short. However, in my 
observations, most instances of feeds using HTML in titles is invalid or 
unambiguous, so adding this feature wouldn't help. For example, RSS 
never allows HTML in titles, though many feeds try anyway.

Atom does allow it if the type attribute is set to "html or "xhtml", and 
Elfeed could check for this attribute and process the title accordingly. 
As you noticed in your code, this is complicated by the fact that no 
HTML parser may be available, and so Elfeed may not always be able to 
parse it. So I'd prefer to just use a simple routine that reliably 
decodes entities and strips all HTML tags, since titles can't be 
meaningfully styled in Elfeed's interface anyway.

The proper place to do this would be in elfeed-entries-from-atom.

ahacop:
Thanks for the pointer to `elfeed-entries-from-atom`.

Here's an implementation that neither requires libxml nor creates thousands of dom representations and temporary buffers. On the other hand, my regexs are probably incomplete or just plain wrong. 😄 

```elisp
(defun ah/decode-html/compose (&rest funs)
  "Return function composed of FUNS. Source: https://nullprogram.com/blog/2010/11/15/"
  (lexical-let ((lex-funs funs))
    (lambda (&rest args)
      (reduce 'funcall (butlast lex-funs)
              :from-end t
              :initial-value (apply (car (last lex-funs)) args)))))

(defun ah/decode-html/entity->character (entity)
  (gethash entity ah/decode-html/entities-hash entity))

(defun ah/decode-html/strip-html-tags (html)
[...]

PR #443, opened 2021-09-15
Clean HTML tags from title

danielfleischer:
Remove tags of the form `< ... >` and `& ... ;`.

These tags appear sometimes, for example in Arts & Letters Daily feed. 

Based on some code in #365.

### DIFF
--- a/elfeed-lib.el
+++ b/elfeed-lib.el
@@ -79,8 +79,10 @@ Align should be a keyword :left or :right."
 
 (defun elfeed-cleanup (name)
   "Trim trailing and leading spaces and collapse multiple spaces."
-  (let ((trim (replace-regexp-in-string "[\f\n\r\t\v ]+" " " (or name ""))))
-    (replace-regexp-in-string "^ +\\| +$" "" trim)))
+  (let* ((trim (replace-regexp-in-string "[\f\n\r\t\v ]+" " " (or name "")))
+         (trim (replace-regexp-in-string "^ +\\| +$" "" trim))
+         (trim (replace-regexp-in-string "\\(<[^>]+>\\|&[^;]+;\\)" "" trim)))
+    trim))
 
 (defun elfeed-parse-simple-iso-8601 (string)
   "Attempt to parse STRING as a simply formatted ISO 8601 date.


sinic:
Character entities, at least, might constitute an integral part of the title, whether as single characters, or because it's talking about HTML entities. Simply removing them won't do.

My pull request at #452 converts them to characters instead, but only for feeds that declare titles to be in HTML. [The Arts & Letters feed](https://www.aldaily.com/feed/) doesn't, unfortunately. For that one, you could adapt [the workaround](https://github.com/skeeto/elfeed/pull/452#issuecomment-1013759957) I just posted.

minad:
@sinic @danielfleischer Not sure if you're still using Elfeed. What about an approach like this?

~~~ elisp
(defun elfeed-cleanup (str &optional html)
   ;; 1. Perform usual whitespace cleanup
   ;; 2. If html, replace entities/tags, otherwise strip them
)
~~~

Then the argument HTML could be set to t where it makes sense (Atom feeds with html title type).

yantar92:
I have local code fighting the same issue. The way I did it was through `elfeed-search-print-entry-function`. I am not sure if something as low level as `elfeed-cleanup` is the right choice here. IMHO, the issue is rendering, not the feed title text itself. Programmatically, one may want titles to be in the original form.

minad:
> IMHO, the issue is rendering, not the feed title text itself. Programmatically, one may want titles to be in the original form.

I am not sure. For atom feeds there can be html in the title, so we could as well convert during saving without losing anything. For other feeds the entities/tags are entirely disallowed, so it is seems okay to sanitize before saving, since the data is otherwise technically invalid. On the other hand one can take your position and say that the data may always contain entities/tags, and then we should fix it during rendering.

yantar92:

> minad left a comment (emacs-elfeed/elfeed#443)
>
>> IMHO, the issue is rendering, not the feed title text itself. Programmatically, one may want titles to be in the original form.
>
> I am not sure. For atom feeds there can be html in the title, so we could as well convert during saving without losing anything. For other feeds the entities/tags are entirely disallowed, so it is seems okay to sanitize before saving, since the data is otherwise technically invalid. On the other hand one can take your position and say that the data may always contain entities/tags, and then we should fix it during rendering.

For some context, I make use of HTML (including invalid) to render bold,
italic, and other styling in feed titles. Loosing that information will
downgrade the experience. I guess we need to settle on what you need by
"convert". If it is removing the tags entirely, I do not like that idea.

As for technically invalid feeds, RSS is infamous for technically
invalid feeds. IMHO, we simply have to deal with that - it is too
common.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,

minad:
> For some context, I make use of HTML (including invalid) to render bold,
italic, and other styling in feed titles. Loosing that information will
downgrade the experience. I guess we need to settle on what you need by
"convert". If it is removing the tags entirely, I do not like that idea.

Okay. I am only in favor of removing tags entirely if they lead to breakage later on, but otherwise it is okay if we are a little bit lax.

> As for technically invalid feeds, RSS is infamous for technically
invalid feeds. IMHO, we simply have to deal with that - it is too
common.

Sure. Do you have a good example which I could try?

sinic:
Sorry @minad, I'm currently not using elfeed and of little help with my old PR.

minad:
@sinic Thanks for responding! What do you use now? Just curious :)

sinic:
My media diet lets me get away with using no feed reader at all, at least for now.

danielfleischer:
It was 5 years ago, it's still working for me. Can close if you want.

danielfleischer:
I've now ran the updated Elfeed (since you took over) and I don't think this patch is needed as I don't encounter the issue anymore, so I'm closing. Thanks all.

Issue #317, opened 2019-05-02
Elfeed is too slow when when there are a lot of rss list


testinggithub1222:
I have been using elfeed for a while now and build up like 30 to 50 rss list. Whenever I call elfeed-update to update everything, it takes quite a long time and hang emacs during the time. It might be 20mn, i am not sure. It consumes 100% of cpu and 54% of my ram which should be 4gb of 8gb of ram and garbage collection always collect. So, my question is, is there a way to make this run faster without interruption or hang emacs? Is that parsing xml consume that much cpu and took so long period of time? Isn't there async process?

skeeto:
Fetching 30 to 50 feeds should only disrupt Emacs for a couple of 
seconds, so this isn't normal.

The very first thing to check is that Elfeed is using curl to fetch 
feeds. The variable elfeed-use-curl indicates whether curl will be used 
or not, and it will automatically set to t if curl was found when Elfeed 
was loaded. If it's nil, you should install curl and ensure it's on your 
PATH. This has two major benefits:

* url-retrieve isn't nearly as asynchronous as advertised, especially 
for certain platforms, but curl is always fully asynchronous. The 
synchronous parts of url-retrieve will block Emacs, but not make it use 
100% CPU.

* url-queue-retrieve doesn't allow for custom headers, which means 
Elfeed can't take advantage of ETag or If-Modified-Since. These features 
prevent Elfeed from reparsing feeds that haven't changed. (It's also 
much nicer to servers.) Elfeed spends most of its time parsing XML, and 
XML parsing is not done asynchronously, so this is important. If you're 
not using curl, Emacs will spend more time at 100% CPU when fetching.

Also make sure Elfeed has been byte-compiled. This should happen 
automatically if you installed it via package.el.

Since you're seeing such excessive memory use, it sounds like one of the 
feeds you're fetching may be humongous. A casual experiment suggests 
that, on 64-bit computers, the s-expression representation of an RSS 
feed is about four times larger than the XML content. This is due to all 
the pointer overhead (lots of cons cells for each element). But if 
you're seeing 4GB of memory use, then that puts the feed at about 800MB 
(buffer + s-exp = 4GB), which seems unlikely. Even more so for the 8GB 
case.

Take a look at each of the feeds in your list and see if any of them are 
particularly large. Also have Elfeed fetch them one at a time to see 
which one causes the problem. If you can narrow it down to a particular 
feed, I'd like to know about it.

testinggithub1222:
[...]

PR #448, opened 2021-10-25
Add mode-class property where appropriate

mssdvd:
This PR adds the property `mode-class` with value `special` to
`elfeed-search-mode` and `elfeed-show-mode`.

According to the elisp docs every "special" mode should have this
property.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -206,6 +206,8 @@ When live editing the filter, it is bound to :live.")
                                          'face 'elfeed-search-filter-face)))
                ("")))))))
 
+(put 'elfeed-search-mode 'mode-class 'special)
+
 (defun elfeed-search-mode ()
   "Major mode for listing elfeed feed entries.
 \\{elfeed-search-mode-map}"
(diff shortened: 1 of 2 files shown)


mssdvd:
The motivation behind this PR is a performance issue that is triggered when Flycheck is enabled for elfeed-search buffers.

`global-flycheck-mode` enables Flycheck where it makes sense (e.g. not in the minibuffer, fundamental-mode etc.). For special buffers like elfeed-search the only way to let know Flycheck what to do is the property `mode-class`.

Without this patch Emacs freeze for many seconds, sometimes minutes.

jiacai2050:
Following #317, With global-flycheck-mode on, my Emacs hang 2 minutes when update feeds.  After disable it, Emacs only hang a few seconds. 
Snip20220505_23

@skeeto Hope you can merge this.

jiacai2050:
```
(setq flycheck-global-modes '(not . (elfeed-search-mode)))
```

For now, users can use this config to disable flycheck in elfeed-search-mode.

minad:
Fixed via https://github.com/skeeto/elfeed/commit/a08867a13982413745f68c3721c506e874d7f1f6

PR #452, opened 2022-01-15
Handle character entities in HTML titles

sinic:
There has been some discussion about this already at issue #365 and #69. A general solution should also do something about HTML tags, but as a stopgap measure I suggest simply replacing (a subset) of the HTML character entities by the respective characters if the title of an Atom entry has an HTML type.

While technically not complete, this solves all issues for my current selection of feeds, and should be strictly more correct than doing nothing.


sinic:
Character entities in titles without HTML type are still a problem, of course. Users of such feeds might want to tag them as `broken` and do something like the following:
```elisp
(defun fix-entities (tag entry)
  (when (elfeed-tagged-p tag entry)
    (elfeed-untag entry tag)
    (setf (elfeed-meta entry :title) (xml-substitute-special
                                      (elfeed-entry-title entry)))))
(add-hook 'elfeed-new-entry-hook (apply-partially #'fix-entities 'broken))
```

PR #457, opened 2022-04-13
Add hook to customize how `elfeed-db-gc` finds elfeed-refs

### LINKED ISSUE #376: content gets overridden on elfeed-search-fetch
I've added a function to `elfeed-new-entry-hook` that adds some metadata and overrides the default content.  If I run `elfeed-search-fetch` (press G) some time later it resets the content of all already existing entries to it's original state (though as expected the things I've set in meta stay the same).

I'm not sure if this is intended behavior.

If so what would be the best way to mitigate this?

zabe40:
This pull request allows users and package authors to extend elfeed's GC mechanism to store elfeed-refs anywhere, not only the content slot of an elfeed entry in the database.

Fixes #376.


zabe40:
Forgot to mention, but all the tests pass.

zabe40:
Hey, @skeeto friendly ping for a merge or otherwise some feedback. I'm nearly ready to call [my own package](https://github.com/zabe40/elfeed-time/) finished, and if at all possible I'd like to avoid recommending that users advise any of elfeed's functions.

gnomon-:
I'm keen to see this merged so [elfeed-paywall can adopt an improved storage model for deeply-fetched content](https://todo.sr.ht/~johnhamelink/elfeed-paywall/1)!

minad:
The problem with this approach is that entries could still get lost if the database is loaded without the `elfeed-db-gc-trace-functions` correctly initialized. What do you think of the alternative of using a separate meta field which contains additional references, e.g., `:extra-refs`? Would that work for your use cases?

minad:
Another simpler option could be to scan the meta plist for `elfeed-ref` values. It wouldn't scan the data structures deeply, and only look at the values directly.

zabe40:
These are both good ideas! Either one would work for me, since I'm just storing one or two additional `elfeed-ref`s per entry. I think I would lean towards scanning the plist for refs instead of the `:extra-refs' field, since that might be a bit safer so there's not a worry of different packages overwriting each other in the one list. 
As a bonus, this would mean that I wouldn't have to change [how my package currently stores the refs](https://github.com/zabe40/elfeed-time/blob/main/elfeed-time.el#L650), and it looks like this also aligns with [how elfeed-tube keeps its refs](https://github.com/karthink/elfeed-tube/blob/2d7251d8363661f7bae3fd6b27ef88edc65d25ed/elfeed-tube.el#L309).

minad:
@zabe40 Closing in favor of #579. Can you please take a look and check if it works for your use case? The tracer recursively scans lists, but no other objects like records or vectors. Both entry and feed metadata are checked.

PR #461, opened 2022-06-26
Fix curl args value by adding quotes around

snowman:
* args value may include space(s)


skeeto:
Thanks for your patch. However, Elfeed builds the `argv` array directly, and arguments are not passed through a shell. Shell quoting is not only unnecessary, it leads to incorrect results. Even if quoting *was* necessary, this particular change would be insufficient anyway (see `shell-quote-argument`). It's actually rather nice of Emacs not to force all process creation through a shell — despite my [unhappiness with its limitations Windows](https://nullprogram.com/blog/2022/02/18/) — since it means we simply don't need to worry about quoting.

You can see this for yourself with a quick test. If I run a fake webserver with netcat so I can see the raw request:

    $ nc -lp1234

Then apply your patch, build+start a clean test environment with `make virtual`, and `M-x elfeed-add-feed` with `localhost:1234/`, I get this from netcat:

    GET / HTTP/1.1
    Host: localhost:1234
    User-Agent: curl/7.74.0
    Accept: */*
    Accept-Encoding: deflate, gzip, br
    'User-Agent: Emacs Elfeed 3.4.1'

Note the single quotes, and the failure in overriding `user-agent`.

snowman:
Thank you for the detailed explanation, it really helps, i didn't realize it :heart:

Issue #476, opened 2022-12-05
elfeed-show-refresh--mail-style base URL is incorrect for kottke.org


kamidon:
I subscribe to the RSS feed of kottke.org at https://feeds.kottke.org/main. However, all the article links are at https://kottke.org/. The `elfeed-show-refresh--mail-style` function computes the `base` URL that it uses to insert HTML into the `elfeed-show` buffer from the `feed` URL. As a result images in the article with relative paths fail to render because they are pulled from `https://feeds.kottke.org/<image path>` instead of `https://kottke.org/<image path>`.

I have worked around the problem by temporarily redefining that function to use the link URL instead of the feed URL. This fixes the problem for kottke.org and I haven't noticed any other problems for other feeds yet. However, I'm unsure whether there are potential issues that I haven't run into yet.

kamidon:
One potential concern that I have is that link URLs could theoretically pointing anywhere, not just at the site I originally configured. It would be nice to lock that down more tightly somehow. Perhaps having an `elfeed-entry-base-url` property for feeds that gets used in place of calculating the base URL from the feed URL if configured.

kamidon:
I looked at this a bit more today. I was thinking of changing `elfeed-show-refresh--mail-style` to check if the metadata on the feed has a `:entry-base-url` plist entry and use that in place of the current base computed from the feed URL if present. Would a change like this be acceptable?

I'm an [elfeed-org](https://github.com/remyhonig/elfeed-org) user as well so if this change would be acceptable I'll also submit a PR to elfeed-org to allow feed metadata to be specified, probably using a property on the feed entries.

kamidon:
Given the proposed commit I made hasn't had an enthusiastic reception I've resorted to solving this a different way for now. I used `:around` advice on `elfeed-compute-base` that intercepts the URL being passed in, matches against it, and allows it to be replaced. If anyone arrives here and is interested in this workaround, the advice function I defined is:

```
 (defvar kea/elfeed-compute-base-url-rewrites nil
   "List of lists of replace-regexp-in-string arguments w/o the actual string")
 (defun kea/elfeed-compute-base-url-rewrite-advice (fn url)
   (mapc (lambda (args)
           (let ((regexp (nth 0 args))
                 (rep (nth 1 args))
                 (optargs (nthcdr 2 args)))
             (setq url (apply #'replace-regexp-in-string regexp rep url optargs))))
         kea/elfeed-compute-base-url-rewrites)
   (funcall fn url))
[...]

PR #479, opened 2023-01-17
Optional :entry-base-url feed property overrides base URL derivation

kamidon:
Bofore this commit, when displaying entries in
'elfeed-show-refresh--mail-style' a base URL would be inserted that was derived from the feed URL by removing any URL path and query parameters. However, some feeds are hosted on a different domain then the entries they reference. For example, the feed for kottke.org is on 'https://feeds.kottke.org/' while the entries are on just 'https://kottke.org/'.

This commit allows the user to override the derived base url for entries of a feed by specify an altenative base URL in the ':entry-base-url' property on the feed.

An alternative implementation that derived the base URL from the entry URL instead of the feed URL was considered but rejected because the entry URLs are controlled by the feed and following the entry URL from the elfeed UI requires a user action while the base URL for displaying embedded content in the feed data would not. With this solution the user explicitly acknowledges an additional location from which they are comfortable loading embedded content.

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -156,7 +156,9 @@ Called without arguments."
          (type (elfeed-entry-content-type elfeed-show-entry))
          (feed (elfeed-entry-feed elfeed-show-entry))
          (feed-title (elfeed-feed-title feed))
-         (base (and feed (elfeed-compute-base (elfeed-feed-url feed)))))
+         (base (and feed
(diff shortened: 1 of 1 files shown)


kamidon:
Resolves issue 476

minad:
I close this for now. I first want to understand if the change is really needed. Discussion continues in #476.

PR #489, opened 2023-07-10
Message "downloaded ..." with no additional formatting

zachlefevre:
URLs with `%` successfully download, but then throws an error for messaging without enough arguments. This fixes it by messaging the URL with no additional formatting.


BBoal:
@zachlefevre why not `(message "%s downloaded" url)` instead?

PR #507, opened 2024-03-26
Fix warnings about unescaped single quotes in docstrings

ZharMeny:
Fixes "docstring has wrong usage of unescaped single quotes" warnings during compilation


skeeto:
Thanks, @ZharMeny. Merged as 5c05a1ea. Two instances were quoted words in 
prose, not code, so I changed them to double quotes instead. I originally 
used single quotes for readability as a docstring. (The byte-compiler is 
wrong to warn about those two instances. Oh well.)

Issue #112, opened 2015-11-24
XML parsing


foudfou:
This feed http://www.heise.de/open/news/news-atom.xml renders void entries in elfeed, although `content` is defined.

skeeto:
I'm surprised this issue hasn't come up yet, since this bug has probably
been around since the beginning. The issue is that this feed uses
embedded XHTML (allowed by Atom). This gets parsed with the rest of the
feed document by the XML parser. When Elfeed builds the entry, it needs
to convert the data structure back into XML so that it can be reparsed
later by shr in elfeed-show.

PR #511, opened 2024-05-31
elfeed--atom-content: use 'summary' if there is no 'content'

blbelew:
Some atom feeds, notably GitLab's project activity feeds (project.atom) include a summary tag but no content tag. Fall back to using the summary as the content if there is no content.

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -205,11 +205,13 @@ This is a workaround for issues in `url-queue-retrieve'."
   (concat "urn:sha1:" (sha1 (format "%s" (or content (float-time))))))
 
 (defun elfeed--atom-content (entry)
-  "Get content string from ENTRY."
-  (let ((content-type (xml-query* (content :type) entry)))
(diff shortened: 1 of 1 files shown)


Thaodan:
I don't think this a good change. If there is no content it should not be shown as such but as summary.

blbelew:
The gitlab project.atom feeds show activity on any project you are subscribed to with gitlab. The way they are formatted, the "content" of the entry is entirely in the summary tag. Elfeed isn't showing that at all right now - so it doesn't display anything at all for the body of the entry. It makes sense that if if there is no content tag, but there is a summary tag, that we at least show that instead of nothing at all.

Thaodan:
I get that but putting the summary into the content tag makes it harder
to detect if there is actual content.
Further what if there's a summary and content?

minad:
Merged. elfeed--atom-content already took content and summary into account, only the xhtml code path was missing.

minad:
This is a follow up of #112, for xhtml summary in addition to xhtml content which was already treated.

PR #521, opened 2024-07-14
Respect `shr-max-image-proportion` when rendering elfeed-entry

Feyorsh:
I noticed that Elfeed would sometimes refuse to scale down images that were too large to be displayed in a given window; after some discussion on the [mailing list](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=71913), I realized this is because Elfeed renders *before* displaying a buffer, where it can't glean any information about window size (because there's no window).

Note that there is still an issue with SVG images not respecting `shr-max-image-proportion`, but that has been fixed in Emacs 30 thanks to @jimporter; that's not Elfeed's problem to fix.

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -214,9 +214,9 @@ The result depends on the value of `elfeed-show-unique-buffers'."
   (let ((buff (get-buffer-create (elfeed-show--buffer-name entry))))
     (with-current-buffer buff
       (elfeed-show-mode)
-      (setq elfeed-show-entry entry)
-      (elfeed-show-refresh))
-    (funcall elfeed-show-entry-switch buff)))
+      (setq elfeed-show-entry entry))
+    (funcall elfeed-show-entry-switch buff)
+    (elfeed-show-refresh)))
 
 (defun elfeed-show-next ()
   "Show the next item in the elfeed-search buffer."


jimporter:
> Note that there is still an issue with SVG images not respecting `shr-max-image-proportion`, but that has been fixed in Emacs 30 thanks to @jimporter; that's not Elfeed's problem to fix.

Emacs 31, actually, though I don't have any particular opposition to making the change in 30 too. (The code has just diverged somewhat between 30 and 31, to fix some issues with zooming images, so I'd need to write a different patch for 30.)

As for the patch, I haven't tried it myself, but from inspection it should do the right thing. Swapping the order around here shouldn't cause any problems I think: so long as there's no redisplay between the two steps (there shouldn't be), everything will be ok and there won't be anything like a [flash of unrendered content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content).

88-percent:
As per the mailing list mentioned, until the updates get pushed out it seems the way to get elfeed to respect `shr-max-image-proportion` is to disable the cache:

``` elisp
(setq shr-ignore-cache t)
```
Let me know if I'm mistaken.

jimporter:
> Let me know if I'm mistaken.

Personally, I'd just [advise](https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html) the function with the new implementation here. Then you can still use the image cache.

minad:
In my tests this issue does not seem to occur anymore. Please ping if you still observe if.

88-percent:
> Please ping if you still observe if.

I'm still getting inconsistent behavior with image resizing if I haven't stored true to `shr-ignore-cache`. On first view the image correctly scales, but if I return to a previously viewed image it doesn't appear to scale. 

OS: Debian 13
emacs: 30.1
elfeed: 20260525.2057
elfeed-goodies: 20221003.1406

minad:
As I've understood @jimporter the bug has been fixed in soon to be released Emacs 31. Therefore I don't see the issue, I guess. Jim, can you confirm this?

We won't add a workaround in Elfeed given the upstream fix, but maybe Jim can post a working advice here.

jimporter:
> As I've understood @jimporter the bug has been fixed in soon to be released Emacs 31. Therefore I don't see the issue, I guess. Jim, can you confirm this?

The Emacs 31 fix is just to make sure SVG images are affected by `shr-max-image-proportion`. My comment over on bug-gnu-emacs about this still applies to Elfeed, based on my most recent tests:

> SHR scales images with respect to the size of the buffer's window. However, by rendering the HTML document with SHR *before* displaying the buffer in a window, it's impossible for Emacs to do that: there's no such window yet! If you swap the 'display-buffer' and 'shr-insert-document' lines though, all should work properly.

(Replace `display-buffer` with `elfeed-show-entry-switch` and `shr-insert-document` with `elfeed-show-refresh` for the Elfeed-specific calls.)

I applied the patch in this PR as advice in my configuration around when this was filed, and it's worked well for me.

---

If you're having trouble reproducing this, here's how I reproduced it:
1. Find an entry with a large image that would be scaled down by SHR
2. Open the entry, see that the image was scaled down
3. Close the entry
4. Open the entry again, see that the image is full-size

Feyorsh:
Fwiw I've been [daily driving](https://raw.githubusercontent.com/Feyorsh/.dotfiles/0d15214eed11aa94a03c50b47a73bf0f27107395/home/config/editor/emacs/patches/elfeed-shr.patch) this patch since I opened this ticket and it's been working perfectly for me.

minad:
@Feyorsh I've added your fix.

PR #524, opened 2024-08-01
fix `elfeed-curl--queue-consolidate` to maintain the same shape spec

SohumB:
under the old implementation, `elfeed-curl--queue-consolidate` wasn't idempotent — if you called it on a consolidated queue, it would continue to add list layers around the consolidated entry. this would then cause `wrong-type-argument` errors later when calling make-process to run curl.

### DIFF
--- a/elfeed-curl.el
+++ b/elfeed-curl.el
@@ -458,11 +458,21 @@ results will not."
             method
             data))))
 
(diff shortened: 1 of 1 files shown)


minad:
Fixed by #555, commit bbb3cac27b0412d80b327b5cfaab83683c96a2a1

PR #527, opened 2024-08-24
elfeed-search: Set mark before showing an entry.

### LINKED ISSUE #526: elfeed-search doesn't set the mark when showing an entry using RET key
Right now, elfeed-search doesn't set the mark when showing an entry using RET key.

That means that back-button cannot go back to the line inside `*elfeed-search*` that was opened.

It would be nicer to set the mark. In addition to back-button being able to go back to the original line, the user can also go back to the original line easier.

daym:
Fixes <https://github.com/skeeto/elfeed/issues/526>.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -847,6 +847,7 @@ browser defined by `browse-url-generic-program'."
   (when (elfeed-entry-p entry)
     (elfeed-untag entry 'unread)
     (elfeed-search-update-entry entry)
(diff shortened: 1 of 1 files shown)


minad:
It seems unusual way to do this. I've not seen other packages using (set-mark-command nil) in similar scenarios. Closing for now. Do you have references where you've seen this pattern before? Pushed an alternative fix.

PR #529, opened 2024-08-26
Normalize queue entries to avoid errors in curl queue consolidation

czan:
This change is attempting to fix the same issue as https://github.com/skeeto/elfeed/pull/524, but in a slightly different way. Rather than maintaining the current shape of queue items, I've attempted to normalise the queue structure to *always* contain a list of URLs.

This has two advantages, to my eyes:
1. it means the queue has a more regular structure, which makes it generally nicer to work with IMO
2. it allows consolidation to work properly when run multiple times, rather than ignoring queue items with a list of URLs (as the code before this change did)

I believe this change will fix the issues reported in https://github.com/skeeto/elfeed/issues/277 (back in 2018!).

### DIFF
--- a/elfeed-curl.el
+++ b/elfeed-curl.el
@@ -410,6 +410,16 @@ DATA is the content to include in the request."
                  for (_ . cb) in elfeed-curl--requests
                  do (run-at-time 0 nil handler buffer cb))))))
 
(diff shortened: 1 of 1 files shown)


minad:
Fixed by #555, commit bbb3cac27b0412d80b327b5cfaab83683c96a2a1

PR #531, opened 2024-10-14
Comply with display-buffer-alist and action

Thaodan:
By using display-buffer we can comply with display-buffer-alist
or display-buffer-base-action/pop-frames.

Includes #540.

### DIFF
--- a/elfeed.el
+++ b/elfeed.el
@@ -561,15 +561,17 @@ called interactively, SAVE is set to t."
 ;; New entry filtering
 
 (cl-defun elfeed-make-tagger
-    (&key feed-title feed-url entry-title entry-link after before
+    (&key feed-title feed-url feed-author entry-title entry-link
+          entry-enclosure entry-content-type after before
           add remove callback)
-  "Create a function that adds or removes tags on matching entries.
+  "Create a function that adds, removes tags or does CALLBACK on matching entries.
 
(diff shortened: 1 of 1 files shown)


Thaodan:
The last two commits refactor article rendering and should probably go into a different PR but please any feedback is welcome.

minad:
@Thaodan Did you consider to set `switch-to-buffer-obey-display-actions` to `t`?

minad:
I close this one, since the `(switch-to-buffer (elfeed-search-buffer))` seems fine to switch to Elfeed. In case one wants to tweak behavior set `switch-to-buffer-obey-display-actions` and `display-buffer-alist`.

Thaodan:

> @Thaodan Did you consider to set `switch-to-buffer-obey-display-actions` to `t`?

Wouldn't that affect every else? I think it's fair to ask to use
display-buffer imho.

A compromise could be to use display-buffer-same-window and then be able
to customize diplay-buffer-alist.


PS: Please give enough time to reply before closing issues!

Thaodan:
I was following the suggestion and trying to get rid of the issues caused by it.

I found these two links which follow my suggestion. I.e. Stefan suggesting to not use switch-to-buffer in lisp code:
- https://emacs.stackexchange.com/questions/27712/switch-to-buffer-vs-pop-to-buffer-same-window
- https://debbugs.gnu.org/cgi/bugreport.cgi?bug=28645#25

PR #543, opened 2025-03-08
Stop requiring other tests

manphiz:
ERT since Emacs 29 disallows tests with the same name.  This can also happen when requiring other tests that have 'ert-deftest' and the required source is run before the requiring source, in which case the required source is run twice and all its 'ert-deftest' are considered reusing test names.  See also bug#66782[1].

This patch removes all requires on other test units, and remove all provides in test unit to prevent such error.  Currently this doesn't cause any issues.  In future, it would be better to put utility variable and functions in a separate helper unit to be able to require.

[1] https://debbugs.gnu.org/cgi/bugreport.cgi?bug=66782

### DIFF
--- a/tests/elfeed-curl-tests.el
+++ b/tests/elfeed-curl-tests.el
@@ -13,6 +13,4 @@
     (cl-loop for (type . url) in table
              do (should (eq (elfeed-curl--protocol-type url) type)))))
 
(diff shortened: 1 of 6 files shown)


manphiz:
Looks like make test only runs the central test file to avoid such issue.  False alarm.

PR #546, opened 2025-06-22
elfeed-web: add charset=utf-8 when sending entry contents

fuzy112:
This fixes the problem where CJK contents cannot be displayed.


fuzy112:
Closing since there seems no interest to merge this.

PR #562, opened 2026-04-20
Add faces for elfeed-entry buffer

bormoge:
This commit adds faces to allow customization for the elfeed-entry buffer without needing to modify the `messages.el` faces. These new faces inherit from the `messages.el` faces to allow backward compatibility with already existing setups.

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -10,12 +10,42 @@
 
 (eval-when-compile (require 'subr-x))
 (require 'shr)
-(require 'message) ; faces
+(require 'font-lock) ; faces
 
 (require 'elfeed)
 (require 'elfeed-db)
 (require 'elfeed-lib)
 (require 'elfeed-search)
+
+(defface elfeed-show-entry-header-face
+  '((t :inherit font-lock-keyword-face))
+  "Face for showing headers in the elfeed-entry buffer."
+  :group 'elfeed)
+
+(defface elfeed-show-entry-body-title-face
+  '((t :weight bold :inherit font-lock-string-face))
+  "Face for showing the title name in the elfeed-entry buffer."
+  :group 'elfeed)
+
+(defface elfeed-show-entry-body-author-face
+  '((t :weight bold :inherit font-lock-string-face))
+  "Face for showing the author name in the elfeed-entry buffer."
+  :group 'elfeed)
+
+(defface elfeed-show-entry-body-date-face
+  '((t :inherit font-lock-string-face))
+  "Face for showing the date in the elfeed-entry buffer."
+  :group 'elfeed)
+
+(defface elfeed-show-entry-body-feed-face
+  '((t :inherit font-lock-string-face))
+  "Face for showing the feed name in the elfeed-entry buffer."
+  :group 'elfeed)
+
+(defface elfeed-show-entry-body-tags-face
+  '((t :inherit font-lock-string-face))
+  "Face for showing the tag names in the elfeed-entry buffer."
+  :group 'elfeed)
 
 (defcustom elfeed-show-truncate-long-urls t
   "When non-nil, use an ellipsis to shorten very long displayed URLs."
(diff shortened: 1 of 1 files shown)


minad:
This looks like a good change to me, however I wonder if we should completely decouple from the message faces.

bormoge:
Let me give you my reasoning for this pull request. If it doesn't convince you feel free to close it.

I am the maintainer of the guava-themes package. This change allows me to do two things:

1. Control exactly which faces of elfeed are changed.
2. Avoid "side effects", i.e. if I change the message.el face maybe I will also change gnus faces, or notmuch faces, or mu4e faces; I honestly wouldn't know if changing the message faces would affect those packages, and if so, I also wouldn't know how they're displayed now either because I don't use those packages.

(By the way congratulations for becoming the new elfeed maintainer).

minad:
Sure, I had understood the motivation. But I wonder if we should even avoid using the message faces altogether. This package is not really related to message.el, so I would prefer if we don't even load it. Could you adapt the PR to instead rely only on the basic font locking faces, please?

bormoge:
I have replaced the message faces with font-lock faces. Here are two screenshots using vanilla Emacs theme so you can compare.

Old:
old

New:
new

minad:
Thanks. Could you please inherit from the same face for the different fields to give a similar look as the old one, only with different colors? Themes can adjust this for consistency with their overall style, more or less colorful.

Also - can we reduce the number of new faces? It seems a little bit fine-grained to me. Introducing 12 new faces at once is a lot. Think about the other theme authors!

bormoge:
Ok before I push any more changes let me show you some screenshots, you tell me which one is the one you like the most, and I'll push that one. And yes, these are all the blues I found in the font lock faces (in the default Emacs theme). 
I'll describe them with a number, the face on the right, and the face on the left.

Edit: I noticed you edited your message. Duly noted. I'll reduce the number of elfeed-show-entry-header-* to one (those are the left ones, the "Title" "Author" "Date", etc.)

1. Left: constant, right: builtin
constant-builtin
</br>

2. Left: function-name, right: builtin
function-builtin
</br>

3. Left: builtin, right: function-name
builtin-function
</br>

4. Left: builtin, right: constant
builtin-constant
</br>

5. Builtin
builtin
</br>

6. Constant
constant


7. Function-name
function-name

minad:
> Ok before I push any more changes let me show you some screenshots, you tell me which one is the one you like the most, and I'll push that one.

Okay :)

I want to use approximately semantic faces. We should not optimize for the perfect look with the default theme. This means font-lock-keyword-face and for example font-lock-string-face. If you have a better idea than font-lock-string-face, let me know.

> Edit: I noticed you edited your message. Duly noted. I'll reduce the number of elfeed-show-entry-header-* to one (those are the left ones, the "Title" "Author" "Date", etc.)

Thanks!

yantar92:

>> Ok before I push any more changes let me show you some screenshots, you tell me which one is the one you like the most, and I'll push that one.
>
> Okay :)
>
> I want to use approximately semantic faces. We should not optimize for the perfect look with the default theme. This means font-lock-keyword-face and for example font-lock-string-face. If you have a better idea than font-lock-string-face, let me know.

Let me allow myself pinging @protesilaos.
He is probably the most knowledgeable when it comes to semantic faces.

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,

bormoge:
I'm a little bit occupied right now so I'll come back in a few hours.

> This means font-lock-keyword-face and for example font-lock-string-face.

I'll push this just in case that ends up being how you want to proceed, and if not I'll modify it once I come back.

minad:
@bormoge Thanks, looks good. We can improve this later if new ideas come up.

minad:
@protesilaos Thank you so much for already updating the faces in your Modus themes!

protesilaos:
You are welcome @minad! By the time I read the ping from @yantar92 I already had a pull request waiting from @Stebalien plus a patch for the same issue in my inbox (I was out of home for a few hours yesterday, doing some work at a local tavern).

People are probably excited with this new development to Elfeed---I know I am! It is a wonderful package and I thank all of you for maintaining it!

minad:
@protesilaos I shortened the names of the faces again such that they only carry the prefix of the feature `elfeed-show-*`, not `elfeed-show-entry-*`.

protesilaos:
Done, thank you!

Issue #484, opened 2023-04-07
Umlauts are not properly rendered in elfeed buffer


jhartotr:
`M-x elfeed RET` shows the following

![image](https://user-images.githubusercontent.com/22501936/230561574-3e2a3051-f965-4e73-b40e-1a1962775cee.png)

Why are umlauts not properly rendered?

`C-u C-x =` on a properly rendered character shows:

![image](https://user-images.githubusercontent.com/22501936/230562091-8f207acb-752c-47f2-b6bd-5997efe08f36.png)

`C-u C-x =` on a wrongly rendered character shows (e.g. `\303`) shows:

![image](https://user-images.githubusercontent.com/22501936/230562344-b2cf5916-b8ac-4dbc-9b19-3a637586aba2.png)

Only elfeed is unable to display the umlauts properly. Gnus, eww, circe, and other emacs packages show the umlauts properly on my system.

I'm using

`GNU Emacs 29.0.60 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.34, cairo version 1.16.0) of 2023-04-06`

and

`elfeed-version` is `3.4.1`

butter-radish:
Those are CSS escapes in the feed titles, right?

Elfeed treats all feed titles as literal strings at the moment, because processing them as HTML is complicated (see #365; RSS feed titles are always supposed to be literal, though, so if this is an RSS and not an Atom feed then it's an incorrectly formatted one)

There are patches that strip away or replace HTML tags in the titles, though (see #452). I suppose you could fashion a similar hack for fixing your umlauts for you, if there are only a few different (escapes for) umlaut characters.

Thaodan:
> Those are CSS escapes in the feed titles, right?

Those a C UTF-8 escape characters. decode-coding-string was enough for me to fix these cases. Using builtin-methods should be the best way of fixing these.

minad:
I tried to reproduce the encoding problem with a few feeds (Tagesschau, Spiegel, Zeit) and umlauts seem to render correctly for me. @Thaodan Do you have some insight on why we're getting a wrong encoding here? Does this happen only on old entries which were encoded as iso-8859-1?

minad:
Closing this old issue for now. I would appreciate a recipe which demonstrates the issue.

Thaodan:
[...]

PR #564, opened 2026-04-26
Decode encoded strings in entry titles. Fixes skeeto/elfeed#484

Thaodan:
Includes #531 and #540.

Feel free to merge in any order.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -443,7 +443,12 @@ The customization `elfeed-search-date-format' sets the formatting."
   (let* ((tags (elfeed-entry-tags entry))
          (date-float (elfeed-entry-date entry))
          (date-str (elfeed-search-format-date date-float))
-         (title (or (elfeed-meta--title entry) ""))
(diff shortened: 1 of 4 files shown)


minad:
I see that there is an encoding issue, but it looks as if the fix is at the wrong place. Shouldn't we decode the titles earlier?

Thaodan:

> I see that there is an encoding issue, but it looks as if the fix is
> at the wrong place. Shouldn't we decode the titles earlier?

Where do you think should the change go?

minad:
No idea yet. But it does not look right to replicate the decoding at multiple places. Maybe it should happen when adding entries to the database.

Thaodan:
> No idea yet. But it does not look right to replicate the decoding at multiple places. Maybe it should happen when adding entries to the database.

Hm I that. Depends on how the data should be saved, decoded or encoded.

minad:
Closing per the discussion in #484

Issue #95, opened 2015-07-06
End of file during parsing


smythp:
On installing Elfeed (either manually or through MELPA) and running M-x elfeed, I get the message "End of file during parsing" in the minibuffer. _Messages_ says it's a problem with elfeed-db-load, but not seeing any unbalanced parens or anything in that function. Any ideas what could be causing this?

smythp:
The index file was corrupted. Deleted it and now Elfeed is working.

danlamanna:
I also encountered this and noticed my index file was size 0. I'm not sure how/why this happened but I'll keep an eye out in the future.

skeeto:
If your write-region-inhibit-fsync is non-nil, then this is a possible
outcome from a system crash or power failure. Since Emacs doesn't use
"fcntl(F_FULLFSYNC)" I believe it's effectively always non-nil for
macOS.

skeeto:
Closing since this is an old issue and a lot has changed.

FrostyX:
This issue happened to me years later (Jan 2023) and removing `~/.elfeed/index` didn't help. However, thank you for the tip @smythp because removing `~/.emacs.d/var/elfeed/db/index` worked.

smythp:
Glad my issue and comment from...eight years ago, yikes...was a help. Have an Emacs-filled 2023.

Issue #80, opened 2015-03-15
Is there a way to recreate the index?


Fuco1:
I closed my emacs now, and after reopening elfeed only thinks I have ~450 things in my index.  When I count all the files in the `data` directory it comes to about 7000.  Is there some way I could recreate the index from scratch?  This means I'm going to loose the read/unread state, but having the search working is more important.

skeeto:
You're seeing garbage in your .elfeed/ directory, that's all. Try
running M-x elfeed-db-compact to compact/optimize the database and
remove all those files. In the end you'll be left with 3 files no matter
how many entries you have. Elfeed tries to run the garbage collector
when Emacs closes to clean these up (e.g. elfeed-db-gc-safe in
kill-emacs-hook).

Why is this? Any time an entry's content changes between updates
(example: YouTube Atom feeds include a view count), it generates a new
"blob" under the data/ subdirectory, so a single entry may have
generated any number of individual files, but at most one is ever in use
per entry. Entries will share blobs if they have the exact same content.

Fuco1:
Well I get about 40 feeds a day, for the past half a year that gives at least 7k items, which about matches the amount of files in the `data` directory.  The index file is now about 200kb, before it used to be 20 megabytes.  

Suppose I deleted the index file, is there a way to rebuild it?  Because I think this is the situation I'm at.

skeeto:
Hmmm, I guess something about the index got corrupted at some point. All
of the significant data is stored in that index file, so the information
is already lost if you're seeing such a big reduction in size. For
comparison to your numbers, at 20k entries, the index on my current
machine is 12MB.

If you're going to delete the index and start over, make sure you have
Emacs closed when you do it. If not, it will restore the index from its
memory when it shuts down. You may as well delete the entire .elfeed/
directory if you go that route, because the data/ subdirectory won't be
relevant anymore either. Elfeed will start a fresh database next time
you use it.

Fuco1:
Out of curiosity, do you version your index file somewhere?

It's not all that important information, but it's nice to have a way to go back and search for something when you need it.

skeeto:
Elfeed doesn't do any sort of versioning of the index. It relies on
Emacs' use of fsync+rename (write-region-inhibit-fsync) to prevent data
loss in the index in the event of a system failure mid-write, so you
should be safe from that. Beyond that you'll have to make the index part
of your normal backups, since there's no way for Elfeed protect it from
a disk failure.

[...]

PR #569, opened 2026-05-03
elfeed-db-save: write index atomically

Stebalien:
Write the elfeed index atomically to avoid corruption. See #95, #80, etc.

### DIFF
--- a/elfeed-db.el
+++ b/elfeed-db.el
@@ -275,8 +275,14 @@ The FEED-OR-ID may be a feed struct or a feed ID (url)."
   (elfeed-db-ensure)
   (setf elfeed-db (plist-put elfeed-db :version elfeed-db-version))
   (mkdir elfeed-db-directory t)
-  (let ((coding-system-for-write 'utf-8))
-    (with-temp-file (expand-file-name "index" elfeed-db-directory)
+  (let* ((coding-system-for-write 'utf-8)
+         (dest (expand-file-name "index" elfeed-db-directory))
+         (temp (concat dest "-tmp"))
+         (write-region-inhibit-fsync nil))
+    ;; We write to a temporary file and rename to avoid corrupting the database
+    ;; on crash. `file-precious-flag' is insufficient as it only works for
+    ;; `save-buffer'.
+    (with-temp-file temp
       (let ((standard-output (current-buffer))
             (print-level nil)
             (print-length nil)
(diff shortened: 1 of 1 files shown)


Stebalien:
Not a replacement for backups, but losing one's index is annoying.

minad:
Makes sense. Should elfeed-db-pack get a similar treatment?

Stebalien:
It should, but it's currently written in two stages: archive then index.

The simplest "atomic" solution is to combine them into a single file (putting the index at the end of the archive). Is that acceptable?

minad:
Maybe we should tolerate a bit of non-atomicity? First save the two temp files, then rename both of them as commit step. It would still be better than the status quo, right?

Stebalien:
How about making the compaction entire process atomic? Instead of compacting then deleting each file we put in the archive:

1. Write the archive/index into a new `data.new` directory.
2. Rename the old `data` directory to `data.old` (there's no way to atomically delete a directory).
3. Rename `data.new` to `data`.
4. Delete `data.old`.

On start:

1. If `data.old` exists, delete it.
2. If `data` doesn't exist but `data.new` does, rename `data.new` to `data`.
3. If both `data.new` and `data` exist, delete `data.new`.

minad:
Sounds good to me!

minad:
I merged this part for now. Regarding the improved compaction, we can work on that later. Feel free to create a PR in case you have something ready. One small problem with the two data.old and data.new directories is that the data.old directory also contains some other unreferenced files, which would be lost by packing, like combining gc+pack.

Stebalien:
Thanks! I started working on it but shelved it for a bit.

> One small problem with the two data.old and data.new directories is that the data.old directory also contains some other unreferenced files

I'm pretty sure any unreferenced files are leftover cruft, right? It looks like `elfeed-db-gc` deletes everything not referenced by the database.

minad:
> I'm pretty sure any unreferenced files are leftover cruft, right? It looks like elfeed-db-gc deletes everything not referenced by the database.

Yes, I believe so. I just wasn't 100% sure if something could get lost, and I found it conceptually a bit less clear if packing implies a direct garbage collection.

Therefore I added the naive implementation via temporary files for now in https://github.com/emacs-elfeed/elfeed/commit/92050986b3e2e28703058648ae884a789036732e. This should reduce the risk a bit.

But of course, I am open to adding the better, fully atomic implementation.

Stebalien:
> Therefore I added the naive implementation via temporary files for now in https://github.com/emacs-elfeed/elfeed/commit/92050986b3e2e28703058648ae884a789036732e. This should reduce the risk a bit.

Yeah, that makes sense. I'll likely get around to this at some point, but the priority has gone from "my db was corrupted" to "my db could theoretically be corrupted again".

PR #575, opened 2026-05-08
Improve spacing alignment in elfeed-search buffer

fuzy112:
Use display property with :align-to to ensure consistent spacing between date and title columns, preventing misalignment issues with variable-width content.

Before the patch: 

Screenshot from 2026-05-09 00-20-58

After the patch:

Screenshot from 2026-05-09 00-20-27


minad:
Thank you, @fuzy112. Merged with small changes in https://github.com/emacs-elfeed/elfeed/commit/b92434b0b18b94402cb6dc1cbbcf92704f2e402e. See also the commits before that which adjust `elfeed-format-column` and `elfeed-search-format-date` to rely on `string-width`.

PR #579, opened 2026-05-10
elfeed-db-gc: Scan entry and feed meta data for elfeed-refs

### LINKED ISSUE #376: content gets overridden on elfeed-search-fetch

minad:
Lists are scanned recursively. Other objects are not scanned.

Fixes #376 and #457

### DIFF
--- a/elfeed-db.el
+++ b/elfeed-db.el
@@ -408,7 +408,12 @@ Runs `elfeed-db-unload-hook' after unloading the database."
 
 (defun elfeed-meta (thing key &optional default)
   "Access metadata for THING (entry, feed) under KEY.
-Return DEFAULT if unavailable."
+Return DEFAULT if unavailable.  During `elfeed-db-gc' and
+`elfeed-db-pack', metadata values will be scanned for `elfeed-ref'
+objects, such that references in metadata will be kept alive.  Note that
+only list data structures will be scanned (e.g., cons, list, alist,
+plist).  The data structures must not be cyclic since this is not
+supported by the database format."
   (or (plist-get (elfeed-meta--plist thing) key)
       default))
 
@@ -536,20 +541,42 @@ Return DEFAULT if unavailable."
                  (remhash id elfeed-db-feeds)))
              elfeed-db-feeds)))
 
+(defun elfeed-db--scan-1 (cb obj)
+  "Scan OBJ for `elfeed-ref' references and call CB for each reference."
+  ;; Written out as a loop to scan lists efficiently since Elisp lacks TCO.
+  ;; Note that cyclic objects are not supported by the database format.
+  (while (cond
+          ((elfeed-ref-p obj)
+           (funcall cb obj)
+           nil)
+          ((consp obj)
+           (elfeed-db--scan-1 cb (car obj))
+           (setf obj (cdr obj))))))
+
+(defun elfeed-db--scan (cb)
+  "Scan database for `elfeed-ref' references and call CB for each reference."
+  (let ((feeds (make-hash-table :test #'equal)))
+    (elfeed-db-visit (entry)
+      ;; Search for references in content, entry metadata and feed metadata
+      (elfeed-db--scan-1 cb (elfeed-entry-content entry))
+      (elfeed-db--scan-1 cb (elfeed-entry-meta entry))
+      ;; Only scan feed metadata once. Keep track of feeds in feeds.
+      (unless (gethash (elfeed-entry-feed-id entry) feeds)
+        (setf (gethash (elfeed-entry-feed-id entry) feeds) t)
+        (elfeed-db--scan-1 cb (elfeed-feed-meta (elfeed-entry-feed entry)))))))
+
 (defun elfeed-db-gc (&optional stats-p)
   "Clean up unused content from the content database.
 If STATS-P is true, return the space cleared in bytes."
   (elfeed-db-gc-empty-feeds)
   (let* ((data (expand-file-name "data" elfeed-db-directory))
          (dirs (directory-files data t "\\`[0-9a-z]\\{2\\}\\'"))
-         (ids (cl-mapcan (lambda (d) (directory-files d nil nil t)) dirs))
-         (table (make-hash-table :test 'equal)))
+         (ids (mapcan (lambda (d) (directory-files d nil nil t)) dirs))
+         (table (make-hash-table :test #'equal)))
     (dolist (id ids)
       (setf (gethash id table) nil))
-    (elfeed-db-visit (entry)
-      (let ((content (elfeed-entry-content entry)))
-        (when (elfeed-ref-p content)
-          (setf (gethash (elfeed-ref-id content) table) t))))
+    (elfeed-db--scan
+     (lambda (ref) (setf (gethash (elfeed-ref-id ref) table) t)))
     (cl-loop for id hash-keys of table using (hash-value used)
              for used-p = (or used (member id '("." "..")))
              when (and (not used-p) stats-p)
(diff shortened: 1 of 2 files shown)


minad:
@zabe40 @gnomon- Please check if this works for your use cases.

minad:
@yantar92 Could you please give this a close look, since it touches the DB and GC code? We don't want any data loss.

yantar92:
Does not look like there will be data loss. But I have comments on the code.

elfeed-db.el
```diff
@@ -598,21 +598,41 @@ Return DEFAULT if unavailable."
...
+VISITED in order to avoid scanning objects multiple times, which matters
+in particular if objects have cycles."
+  ;; Written out as a loop to scan lists efficiently since Elisp lacks TCO.
+  (while (cond
+          ((elfeed-ref-p obj)
+           (setf (gethash (elfeed-ref-id obj) refs) t)
+           nil)
+          ((consp obj)
+           (unless (gethash obj visited)
+             (setf (gethash obj visited) t)
```
yantar92: Do remember that gethash is not reliable for deeply nested lists. Hashes are only computed up to certain nesting depth. If two different lists are differenent beyond that nesting, hashes will remain identical.

elfeed-db.el
```diff
@@ -598,21 +598,41 @@ Return DEFAULT if unavailable."
...
+  "Trace OBJ for `elfeed-ref' references.
+Found references are stored in REFS.  Visited objects are stored in
+VISITED in order to avoid scanning objects multiple times, which matters
+in particular if objects have cycles."
+  ;; Written out as a loop to scan lists efficiently since Elisp lacks TCO.
+  (while (cond
+          ((elfeed-ref-p obj)
+           (setf (gethash (elfeed-ref-id obj) refs) t)
+           nil)
+          ((consp obj)
```
yantar92, line 552: I can see that consp recursion is specifically aiming for elfeed-entry-meta plists. But note that someone using a hash table or vector inside will run into misses. On the other hand, GC is a rabit hole (e.g. see memory-report.el). Not sure how deep we need to dive into it.

elfeed-db.el
```diff
@@ -598,21 +598,41 @@ Return DEFAULT if unavailable."
...
     (dolist (id ids)
-      (setf (gethash id table) nil))
-    (with-elfeed-db-visit (entry _)
-      (let ((content (elfeed-entry-content entry)))
-        (when (elfeed-ref-p content)
-          (setf (gethash (elfeed-ref-id content) table) t))))
-    (cl-loop for id hash-keys of table using (hash-value used)
+      (setf (gethash id refs) nil))
+    (with-elfeed-db-visit (entry feed)
+      (clrhash visited)
```
yantar92: Do you want to clear visited on every single entry? What about cross-entry references?

elfeed-db.el
```diff
@@ -598,21 +598,41 @@ Return DEFAULT if unavailable."
...
+      ;; Search for references in content, entry metadata and feed metadata
+      (elfeed-db--trace visited refs (elfeed-entry-content entry))
+      (elfeed-db--trace visited refs (elfeed-entry-meta entry))
+      (elfeed-db--trace visited refs (elfeed-feed-meta feed)))
```
yantar92: This will run repeatedly for every entry in the feed.

elfeed-db.el
```diff
@@ -598,21 +598,41 @@ Return DEFAULT if unavailable."
...
-      (setf (gethash id table) nil))
-    (with-elfeed-db-visit (entry _)
-      (let ((content (elfeed-entry-content entry)))
-        (when (elfeed-ref-p content)
-          (setf (gethash (elfeed-ref-id content) table) t))))
-    (cl-loop for id hash-keys of table using (hash-value used)
+      (setf (gethash id refs) nil))
+    (with-elfeed-db-visit (entry feed)
+      (clrhash visited)
+      ;; Search for references in content, entry metadata and feed metadata
```
yantar92: This gotta be documented.

elfeed-db.el
minad: Yes, I want to avoid that the hash table grows too much, since at this point we might get a negative performance impact. Cross entry references are not really allowed, but even if they exist, each entry would be scanned separately.

elfeed-db.el
minad: I know, yes. Perhaps the performance impact of this repeated scan is worse than keeping the larger visited hash?

elfeed-db.el
minad: Indeed. Probably in the docstring of `elfeed-meta`.

elfeed-db.el
minad: The problem is that for lists every cons cell is added to the visited hash to ensure that the scanner does not hang on cyclic objects. This means that the visited hash table might grow a lot. We don't know what people store inside the meta data, so we must guard against cycles.

elfeed-db.el
minad: No, I use `:test #'eq` here, this is a hash on object identity!

elfeed-db.el
minad: Well. We do not support scanning for references at all right now. If we document that we only scan for references in list data structures, it should be okay.

elfeed-db.el
yantar92: Why not simply looping over lists then?

elfeed-db.el
minad: Also if users come and tell us that they store large hash tables or vectors, then we can reconsider. But then, this is malpractice. Larger data should be stored externally via a ref. No large objects should be stored inside the elfeed-entry and elfeed-feed objects. This means lists/alists/plists should suffice.

yantar92: I agree that documenting is the best way to go.

elfeed-db.el
minad: For cycles garbage collection will hang. Since we do not control the metadata we cannot risk that.

elfeed-db.el
minad: Resolved in https://github.com/emacs-elfeed/elfeed/pull/579/commits/0b7ad0bdbf0dc0de9c57a2e7d77bde3c8e160454.

elfeed-db.el
minad: Documented in https://github.com/emacs-elfeed/elfeed/pull/579/commits/0b7ad0bdbf0dc0de9c57a2e7d77bde3c8e160454.

elfeed-db.el
yantar92: Looking into memory-report.el, it does not look like hash table size is an issue to worry about.

yantar92: You are right about loops, of course.

minad: > Looking into memory-report.el, it does not look like hash table size is an issue to worry about.

If we track every single cons cell in the metadata over the entire database it feels wrong. I cannot guarantee that this won't lead to problems, therefore the additional care.

yantar92: At worst, it is a hash for every cons cell. May be some extra memory usage, but it is comparable to what the ordinary GC does frequently anyway.

yantar92: I also tested on my db with 240k records, and nothing catastrophic happens without clrhash. Of course, another story is for people who actually have references to trace, but I do not expect probels in practice.

minad: No, the ordinary GC keeps track of objects via a mark bit. One bit per object, not one hash entry per object.

minad: I improved the implementation. The clrhash stays to keep the table small, and there is an additional hash to avoid rescanning the feeds. I hate the complexity though.

yantar92: Ok. Let's not pedal this then. I am ok to leave the limitation and just say that cross-references are not supported.

minad: Another option mentioned in #457: Treat the metadata list as flat and only scan for refs on the first level.

Basically this:

~~~
(when (proper-list-p metadata-list)
  (dolist (val metadata-list)
    (when (elfeed-ref-p val)
      (mark val)))
~~~

The visited-feed table might still be needed. But we would avoid the visited-cons table. However I wonder if this is way too restricted for users. It would be good to solve this, such that it works in general.

yantar92: That will be weird. Different rules for scanning different things tend to be confusing on the user side.

elfeed-db.el
minad: Fixed by visited-feed hash table.

minad:
@yantar92 I'd appreciate another critical look. I extracted the function `elfeed-db--scan` and use it in both `elfeed-db-gc` and `elfeed-db-pack`. I think it is okay like this, since the details are abstracted away in `elfeed-db--scan`. What do you think?

minad:
One more simplification: Since the database format does not use cycles (print-circles=nil), we can also avoid this here, and get rid of the visited-cons hash table. There is no need to lift the restriction to solve the problem of scanning for elfeed-refs in metadata. This means a large part of our discussion above wasn't actually necessary :/ But well, we end up with simpler code, so I am happy.

yantar92:
> One more simplification: Since the database format does not use cycles (print-circles=nil), we can also avoid this here, and get rid of the visited-cons hash table. There is no need to lift the restriction to solve the problem of scanning for elfeed-refs in metadata. This means a large part of our discussion above wasn't actually necessary :/ But well, we end up with simpler code, so I am happy.

Probably need to make sure that this limitation is written down somewhere.

minad:
> Probably need to make sure that this limitation is written down somewhere.

Yes, this is a documented historical limitation. See the commentary in elfeed-db.el. I've also documented this in the `elfeed-meta` docstring.

We could lift the cycle restriction at some later point, but this may also lead to difficulties if we experiment with alternative databases. I would avoid lifting any restriction as long as there is no explicit need. This PR should only address the references in metadata problem and nothing else.

yantar92:

> I would avoid lifting any restriction as long as there is no explicit
> need

+1

-- 
Ihor Radchenko // yantar92,
Org mode maintainer,

zabe40:
> Please check if this works for your use cases.

Yes this works, thanks so much!

Issue #380, opened 2020-05-14
[RFE] Use a real database for storing data


FrostyX:
Was there any discussion about whether to use a real database for storing data? I am not familiar with `elfeed` internals but I believe it should be doable? I don't propose it to be the default way to store data, the current implementation is much better for that because it doesn't have any external dependencies. However, in some (not-so-special) cases it would be useful to have an access to big guns.

From a performance perspective:

- I use elfeed only for a few weeks (and I love it), so my `~/.elfeed/index` is just fine, but from what I've seen in the issue tracker, it can grow a lot and cause some performance issues.
- I am afraid that text file database is prune to failures when writing (maybe this is well-handled in the code and it is not possible to introduce inconsistencies or unparseable index file)

Additional values:

- It would allow us to easily query data from outside of Emacs. A simple, yet very useful use-case for this is a possibility to implement an indicator for new articles in your window manager's panel. 
- It would allow us to run an unlimited number of complex SQL queries for fetching whatever statistics (I know there is `elfeed-csv-export`, so this may not be so attractive benefit)
- Sharing elfeed data between multiple computers would be easier than ever

Personally, I would like to use elfeed with SQLite but for some use-cases, it would be useful to run it with PostgreSQL or MariaDB, so the implementation should be abstract enough.

skeeto:
Here's overview of the database:

The Elfeed Database
https://nullprogram.com/blog/2013/09/09/

Only a few minor details have changed, so that's essentially still how 
it works. There are a few things I'd do differently now that I have more 
experience: 1) do not depend on cl-struct because it turned out to be 
unreliable, 2) be more flexible about indexing, 3) less denormalization 
in the file representation by doing more work at load time (which is 
actually faster), 4) rely more on print-circle (too immature prior to 
Emacs 24). Basing the index around the avl-tree package, and the content 
database both turned out to be solid ideas.

The database is written durably if write-region-inhibit-fsync is nil 
(the default) and Emacs implements it properly (not possible on all 
platforms), so users need not worry about losing the entire index — the 
actual important part of the database — due to a power failure. The 
database is written in its entirety each time it's saved, which is the 
main factor limiting its scalability. The content database specifically 
disables fsync for writes for performance reasons, but it's also 
considered volatile and unimportant.

One of my early plans was to eventually switch the database over to 
SQLite. I wrote an entire package for that purpose, EmacSQL:

Introducing EmacSQL
https://nullprogram.com/blog/2014/02/06/

[...]

PR #589, opened 2026-05-13
put elfeed-db-* functions behind a vtable

butterfingres:
This allows for making another database implementation for #380.


minad:
Hi! I appreciate the effort, but it is unlikely that we will merge this. We still have to decide how to proceed, see #586. Instead of a vtable implementation we might use cl-defgeneric instead.

minad:
Closing this for now. Discussion continues at #586.

PR #605, opened 2026-05-19
Improve elfeed search hook handling

bormoge:
Hello again.

This pull request contains two commits. 

The first one is just an opinionated refactor to the already existing code. In practice this doesn't actually modify the functionality of Elfeed, but rather it groups (almost) all the used hooks into one function. This should make it easier to manage hooks.

The second commit adds a custom variable to manage separators and, probably more important, it makes it easier to add and remove values on elfeed-search-update-hook in the future. From what I have noticed when using Emacs, hooks are more of a dynamic mechanism, and asking the user to directly interface with them may or may not cause unforeseen effects.

Feel free to accept one, two, or neither commits.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -46,6 +46,11 @@
 The functions may modify the search buffer or add overlays, for example
 `elfeed-search-add-separators'.")
 
+(defcustom elfeed-search-separators t
+  "Add separators to Elfeed entries according to their date."
+  :group 'elfeed
+  :type 'boolean)
+
 (defcustom elfeed-search-update-delay 0.2
   "Delay search buffer updates to avoid redundant redraws.
 The delay is in seconds."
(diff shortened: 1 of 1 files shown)


minad:
Thanks. I moved the minibuffer setup hook. It doesn't make a big difference.

> The second commit adds a custom variable to manage separators and, probably more important, it makes it easier to add and remove values on elfeed-search-update-hook in the future. From what I have noticed when using Emacs, hooks are more of a dynamic mechanism, and asking the user to directly interface with them may or may not cause unforeseen effects.

I don't really agree with this statement. Users should not be afraid to add or remove hooks. For instance user configs usually add mode hooks which enable certain minor modes buffer locally.

bormoge:
Understandable. Thank you minad.

minad:
@bormoge Btw, elfeed-search-separator-date-format can be customized and set to nil.

bormoge:
The way you implemented elfeed-search-separator-date-format is much more flexible. Thank you minad.

PR #615, opened 2026-05-21
Add elfeed-show-entry-date-format option

bormoge:
This commit is simple: it adds a variable that allows users to customize the date shown in the elfeed-entry buffer.


minad:
Thanks. Note that I've shortened the name to `elfeed-show-date-format`. For some reason the prefix `elfeed-show-entry-*` proliferated, but `elfeed-show-*` is actually sufficient. I've also shortened the face names from your PR #562.

PR #616, opened 2026-05-23
Split elfeed-search-print-entry--default

elfehr:
The display of the search buffer is customizable with the variable `elfeed-search-print-entry-function`, but `elfeed-search-print-entry--default` is a monolithic function that inserts everything at once. That makes it difficult to insert custom columns or modify the appearance of a single column without having to copy-paste the whole function.

The columns are almost completely independent, so `elfeed-search-print-entry--default` can be split cleanly. This pull request splits it into one function per column, and `elfeed-search-print-entry--default` simply concatenates and insert the propertized strings. 

The only difference with the original function is that I removed the `:align-to` property from the spaces after the date and the entry title. They are already formatted to have a fixed width, so the property doesn't seem needed, and aligning to a fixed position in the buffer doesn't work well if custom columns are added.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -524,11 +524,14 @@ never show a relative time."
 
 (defun elfeed-search-format-column--date (entry)
   "Format the date information for ENTRY."
-  (let ((date (elfeed-search-format-date (elfeed-entry-date entry))))
-    (elfeed-add-properties date
-                           'face 'elfeed-search-date-face
-                           'mouse-face 'highlight
-                           'follow-link [elfeed-date])))
+  (let ((date (elfeed-search-format-date (elfeed-entry-date entry)))
+        (date-width (cadr elfeed-search-date-format)))
+    (cons
+     (elfeed-add-properties date
+                            'face 'elfeed-search-date-face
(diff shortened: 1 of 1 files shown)


minad:
> The only difference with the original function is that I removed the :align-to property from the spaces after the date and the entry title. They are already formatted to have a fixed width, so the property doesn't seem needed, and aligning to a fixed position in the buffer doesn't work well if custom columns are added.

The space has been added recently to handle variable pitch font and wide characters, so this must be preserved. My suggestion would be that each column function returns a string and a width. Please also test with `variable-pitch-mode` enabled.

minad:
Out of interest, which custom columns do you add?

elfehr:
Sorry, I should have thought about that! Now it also looks the same with `variable-pitch-mode`.

I mostly use elfeed to follow youtube feeds, so I wanted to add a column for the duration of the video. I have a function in `elfeed-new-entry-hook` that queries the duration, and I used to store it as tag, but I figured it was time to make it a proper metadata.

minad:
Thanks! Merged. I made a few more adjustments.

PR #636, opened 2026-06-09
Removed scrolling to the top of buffer when showing an entry

### LINKED ISSUE #635: Emacs bookmarks not navigating to correct position in elfeed entry
When using Emacs bookmarks to navigate to a position in an elfeed entry the pointer ends up in the wrong location. 

Here's an example of my bookmark:

```elisp
("elfeed entry \"2026-06-08 Emacs news\""
 (id "sachachua.com" . "https://sachachua.com/blog/2026/06/2026-06-08-emacs-news/")
 (location . "2026-06-08 Emacs news")
 (position . 6140)
 (handler . elfeed-show-bookmark-handler))
```

Instead of navigating to position 6140 it navigates to position 590.

The culprit is line 270 in elfeed-show.el:

```elisp
      (when-let* ((win (get-buffer-window buffer)))
	(set-window-start win (point-min))))))
```

This was introduced in commit 0228956 and removing these lines fixes the issue. I tested for a bit and I don't see any negative impact of removing this.

I'll submit a PR shortly with this change but there could be side effects to this change, so someone more knowledgeable should comment on the PR and I can try to make whatever change is necessary.

lodenrogue:
Fixes #635

### DIFF
--- a/elfeed-show.el
+++ b/elfeed-show.el
@@ -265,10 +265,7 @@ The result depends on the value of `elfeed-show-unique-buffers'."
       (setq elfeed-show-entry entry))
     (funcall elfeed-show-entry-switch buffer)
     (with-current-buffer buffer
-      (elfeed-show-refresh)
-      ;; Scroll to top after reset
-      (when-let* ((win (get-buffer-window buffer)))
-        (set-window-start win (point-min))))))
+      (elfeed-show-refresh))))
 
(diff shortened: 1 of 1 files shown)


minad:
No, this patch is not correct. Scrolling to the top is necessary when resetting the entry buffer, when moving between entries. I'll think about a better solution.

PR #638, opened 2026-06-15
Add elfeed-search-next-separator

bormoge:
Hello.

This commit adds a function that helps the user traverse separators. It can be bound to a key and used with or without the universal argument.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -1423,6 +1423,43 @@ The update is delayed by `elfeed-search-live-delay'."
     ;; Delete unnecessary separator again if there is only a single one.
     (when (= count 1) (delete-overlay ov))))
 
+;; Functions to traverse separators
+(defun elfeed-search-next-separator (n)
+  "Move cursor to the next N separator.
+Positive N moves forward, negative N moves backward."
+  (interactive "p")
+  (unless (zerop n)
+    (let* ((bol (pos-bol))
+           (forward (> n 0))
+           (candidates
+            (sort
+             (delete-dups
+              (mapcar #'overlay-start
+                      (cl-remove-if-not
+                       (lambda (ov)
+                         (eq (overlay-get ov 'category)
+                             'elfeed-search-separator))
+                       (if forward
+                           (overlays-in bol (point-max))
+                         (overlays-in (point-min) bol)))))
+             #'<))
+           (current-candidate
+            (nth (if forward
+                 (abs n)
+               (1- (abs n)))
+                 (if forward
+                     candidates
+                   (reverse candidates)))))
+      (if current-candidate
+          (goto-char current-candidate)
+        (user-error "No separators left")))))
+
+(defun elfeed-search-previous-separator (n)
+  "Move cursor to the previous N separator.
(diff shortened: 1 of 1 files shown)


minad:
Thanks. Good idea. This should remap forward-paragraph and backward-paragraph.

elfeed-search.el
```diff
@@ -1423,6 +1423,28 @@ The update is delayed by `elfeed-search-live-delay'."
...
     (when (= count 1) (delete-overlay ov))))
 
+;; Function to traverse separators
+(defun elfeed-search-next-separator (n)
+  "Move cursor to the next N separator.
+Positive N moves forward, negative N moves backward."
+  (interactive "p")
+  (unless (zerop n)
+    (let* ((current-pos (point))
+           (separators (sort (overlays-in (point-min) (point-max))
```
minad: You should check for the elfeed-search-separator category here somewhere.

elfeed-search.el
```diff
@@ -1423,6 +1423,28 @@ The update is delayed by `elfeed-search-live-delay'."
...
+                             (lambda (a b)
+                               (< (overlay-start a)
+                                  (overlay-start b)))))
+           (positions (delete-dups
+                       (mapcar #'overlay-start separators)))
+           (candidates (if (> n 0)
+                           (seq-filter (lambda (s) (> s current-pos)) positions)
+                         (reverse
+                          (seq-filter (lambda (s) (< s current-pos)) positions))))
+           (current-candidate (nth (1- (abs n)) candidates)))
```
minad: This is only a stylistic point, but please use cl-lib (e.g. cl-loop instead of seq-* here), for consistency with the other Elfeed code. Not that it really matters, but it seems that the code can be written much more efficiently with fewer intermediate lists.

bormoge:
I've replaced seq-filter with cl-remove-if-not and put a elfeed-search-separator check. I've also modified the order of the code to optimize it a little bit more.

minad:
Okay, thanks! We still need `elfeed-search-previous-separator` which calls `elfeed-search-next-separator`, and then these should remap the aforementioned commands.

elfeed-search.el
```diff
@@ -1423,6 +1423,32 @@ The update is delayed by `elfeed-search-live-delay'."
...
+           (positions
+            (sort
+             (delete-dups
+              (mapcar #'overlay-start
+                      (cl-remove-if-not
+                       (lambda (ov)
+                         (eq (overlay-get ov 'category)
+                             'elfeed-search-separator))
+                       (overlays-in (point-min) (point-max)))))
+             #'<))
```
minad: Here it would be better to obtain the overlays in the ranges `(point-min), (pos-bol)` and `(pos-bol), (point-max)` depending on the direction. Then one wouldn't need the `cl-remove-if-not` below.

bormoge:
I have removed the other two cl-remove-if-not. One detail I noticed is that the backward list of candidates has one more element than the forward list.

minad:
Thanks. I merged this with a few improvements.

PR #640, opened 2026-06-16
Add elfeed-search-set-separator-date-format

bormoge:
This commit adds a function that allows the user to interactively set the separator date format.
This should be especially useful now that the next-separator and previous-separator functions were added to elfeed.

### DIFF
--- a/elfeed-search.el
+++ b/elfeed-search.el
@@ -1404,6 +1404,23 @@ The update is delayed by `elfeed-search-live-delay'."
                  (get nil 'elfeed-search-separator))
              entry)))
 
+(defun elfeed-search-set-separator-date-format (new-format)
+  "Set `elfeed-search-separator-date-format' to NEW-FORMAT.
+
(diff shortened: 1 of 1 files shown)


minad:
Thank you for the proposal. This looks to me like overdoing it, and is probably a better fit for the Elfeed wiki or simply as a personal command in your user configuration. You can already use `set-variable` on arbitrary customizable variables, so we won't need a specific command only for the date format customizable variable.

However I would imagine that a command which allows you to cycle between daily, weekly, monthly date formats could be useful (see also the command `elfeed-search-order-cycle`). Maybe even a general cycle command could be useful which cycles across daily, weekly, monthly, and by-feed separators.

But please let's discuss this first in a separate discussion, since we have to find a way to integrate this nicely in the existing infrastructure.

bormoge:
Sure, I'll open a discussion.

(I didn't know about `elfeed-search-order-cycle`, thanks for mentioning it)

PR #647, opened 2026-07-14
Use header-line-highlight face instead of highlight face for header-line elements

protesilaos:
The `header-line-highlight` was introduced in Emacs 28, while Elfeed is for Emacs 29+.


minad:
Thanks! I didn't know about this face.

PR #657, opened 2026-08-18
Update elfeed-search.el

Tonus1:
Change incf to cl-incf


minad:
This is not correct. Look into Emacs 31 and Compat 31.

